PR #3, opened 2018-10-22
add Makefile

sjenning:


sjenning:
self-merging so I can get automated builds done in CI

PR #5, opened 2018-10-23
change image path to ci built image

sjenning:


sjenning:
Actually we have a :chicken: :egg: situation here.  The image won't be promoted until a PR merges into master for this repo. Push forward!
/hold cancel

PR #6, opened 2018-10-30
Fix descheduler binary path

ravisantoshgudimetla:
/cc @sjenning 

Fixing the binary path


ravisantoshgudimetla:
/lgtm

@sjenning - Some of the fixes are making are so small. So lgtm'ing them myself :).

sjenning:
uuuh, it this correct? https://github.com/openshift/descheduler/blob/master/Dockerfile#L24

PR #7, opened 2018-10-31
Code clean up and RBAC configuration

ravisantoshgudimetla:
- Restructured RBAC with rolebindings so that we don't have to create a roles/clusterroles/rolebindings in operator. 
- Deleted memcached.csv.yaml which was pushed accidentally.

/cc @sjenning


ravisantoshgudimetla:
@sjenning - Merging this as https://github.com/openshift/descheduler-operator/blob/d8734d386751513a80ff1e31ef154a49f957f634/pkg/stub/handler.go#L103, fixes the code to point to correct descheduler image.

PR #12, opened 2018-11-08
Add basic validation for strategies

ravisantoshgudimetla:
Add some basic validation for strategies. As of now, based on string validation that is shoved into configmap.

/cc @sjenning

### DIFF
--- a/pkg/apis/descheduler/v1alpha1/descheduler_types.go
+++ b/pkg/apis/descheduler/v1alpha1/descheduler_types.go
@@ -14,6 +14,11 @@ type DeschedulerSpec struct {
 }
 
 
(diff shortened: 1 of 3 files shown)


sjenning:
can we add some unit tests?

ravisantoshgudimetla:
Sure, I wanted to have e2es first though. Writing unit tests involve a lot of mocking(especially at the controller level). I will add some unit tests tomorrow for some functions.

sjenning:
seems like it wouldn't be hard to turn the validation code block into a function that only took the strategies as an argument and would be easier to unit test. `Reconcile()` could use some decomposition anyway IMHO.

PR #23, opened 2018-12-01
Make e2e run in CI

ravisantoshgudimetla:
This commit 

- Updates Make file and e2e scripts to run locally so that we can test them in CI
- Updates operator-sdk godeps to v0.2.0 which has latest changes and other dependency updates

This script is tested locally and running fine.

The final step here is to provision a 4.0 cluster and run these tests against that cluster to validate e2e is working properly. I am keeping a WIP as of now, I am yet to test it with a CI provisioned 4.0 cluster.

/cc @sjenning


test/e2e/descheduler_test.go
```diff
@@ -152,17 +151,8 @@ func DeschedulerCluster(t *testing.T) {
 		t.Fatalf("failed to initialize cluster resources: %v", err)
 	}
 	t.Log("Initialized cluster resources")
-	namespace, err := ctx.GetNamespace()
```
ravisantoshgudimetla: We don't need this as we are not running as deployment. 
Note to myself: Add a flag which checks for deployments in case someone is interested in running as deployment.

hack/run-e2e.sh
```diff
@@ -0,0 +1,14 @@
+set -o errexit
+set -o nounset
+set -o pipefail
+
+DESCHEDULER_OPERATOR_ROOT=$(dirname "${BASH_SOURCE}")/..
+
+# Get operator-sdk binary.
+# TODO: Make this to download the same version we have in godeps
+wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.2.0/operator-sdk-v0.2.0-x86_64-linux-gnu
```
ravisantoshgudimetla: This is more ci-specific, if someone wants to run locally, there needs to be a way for it. One alternative is to add kubeconfig flag, if nothing provided run in CI mode where we assume kubeconfig is present at a particular path.

hack/run-e2e.sh
```diff
@@ -0,0 +1,14 @@
...
+chmod +x /tmp/operator-sdk
+cd $DESCHEDULER_OPERATOR_ROOT
+kubectl create -f deploy/namespace.yaml --kubeconfig /tmp/admin.kubeconfig
```
ravisantoshgudimetla: Need to check if `/tmp/admin.kubeconfig` is valid location for kubeconfig when running in CI.

PR #24, opened 2018-12-03
Fix e2e tests kubeconfig path

ravisantoshgudimetla:
Wanted to test e2e, so I am removing the kubeconfig now, to test, if we are running e2e


ravisantoshgudimetla:
E2es are passing now. Merging the pr.

PR #28, opened 2018-12-05
Update readme with latest CR changes

ravisantoshgudimetla:
Update readme with latest CR changes.

/cc @sjenning


ravisantoshgudimetla:
Not able to reproduce this locally. 

/retest

PR #41, opened 2019-01-22
Expose image to be managed by descheduler operator

ravisantoshgudimetla:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1666649

As of now, we are not exposing the information related to descheduler image being managed by operator, this PR should help in fixing it.

/cc @sjenning

### DIFF
--- a/pkg/controller/descheduler/descheduler_controller.go
+++ b/pkg/controller/descheduler/descheduler_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"reflect"
 	"strings"
 
 	deschedulerv1alpha1 "github.com/openshift/descheduler-operator/pkg/apis/descheduler/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/controller/descheduler/descheduler_controller.go
```diff
@@ -462,8 +478,7 @@ func (r *ReconcileDescheduler) createCronJob(descheduler *deschedulerv1alpha1.De
 							RestartPolicy:     "Never",
 							Containers: []v1.Container{{
 								Name:  "openshift-descheduler",
-								Image: "registry.svc.ci.openshift.org/openshift/origin-v4.0:descheduler", // TODO: Make this configurable too.
```
ravisantoshgudimetla: @sjenning - Let me know, if you like this way of exposing image within CR, I see that not many people are doing it this way.

pkg/controller/descheduler/descheduler_controller.go
```diff
@@ -26,8 +26,9 @@ import (
 )
 
 const (
-	Running  = "RunningPhase"
-	Updating = "UpdatingPhase"
+	Running      = "RunningPhase"
+	Updating     = "UpdatingPhase"
+	defaultImage = "registry.svc.ci.openshift.org/openshift/origin-v4.0:descheduler"
```
ravisantoshgudimetla: Should this be image coming from quay?

PR #47, opened 2020-01-20
Refactor to use library-go instead of operator-sdk

damemi:
As part of ongoing work toward https://issues.redhat.com/browse/WRKLDS-121, this PR refactors much of the framework code to use our library-go patterns (as well as updating dependencies and switching to go.mod)


damemi:
/hold
Looks green, just want to tidy up a few bits before we merge

PR #48, opened 2020-01-20
Rename references to descheduler-operator package

damemi:
With the repo name change, CI is failing. I have a PR to update the CI here: https://github.com/openshift/release/pull/6776 and this will need to be passed through to our current build


damemi:
This may be a high priority change, as I don't know (yet) if this is breaking CI/release tooling for just this repo or everywhere

damemi:
This also might need a forced merge, since the CI change is cyclically dependent on this change (and vice versa)

damemi:
Also updated this PR to change OWNERS to match kso

PR #60, opened 2020-01-21
Fix file path

kasturinarra:
Fix path of yaml file in oc create commands which are used to deploy operator


ingvagabund:
Deployment instructions have changed.
/close

PR #65, opened 2020-01-28
Bug 1798500: Add nodetaints strategy and refactor configmap string generation to use structs and marshalling

damemi:
This refactors how the configmap policy string is generated to use native descheduler types instead of string construction, and also adds support for the `RemovePodsViolatingNodeTaints` strategy


pkg/operator/target_config_reconciler.go
```diff
@@ -36,6 +38,9 @@ var validStrategies = sets.NewString("duplicates", "interpodantiaffinity", "lown
 // compatibility with 3.11
 var DeschedulerCommand = []string{"/bin/descheduler", "--policy-config-file", "/policy-dir/policy.yaml", "--v", "5"}
 
+type strategyPolicy struct {
```
ingvagabund: No need to add it here when it's removed in the second commit.

PR #66, opened 2020-01-28
Bug 1797745: Setup bindata for resources

damemi:
Currently we're creating the configmap and cronjob manually when we could load in a lot of it from bindata, clean up the code, and also take advantage of library-go resourceapply helpers


damemi:
https://github.com/openshift/library-go/pull/689 would add resourceread/apply methods for CronJob

damemi:
/hold
waiting for `--descheduling-interval` to be merged upstream

PR #67, opened 2020-02-03
Fix CR apiversion and target namespace/name

ingvagabund:
CR is expected to live under `openshift-kube-descheduler-operator/cluster`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,21 +12,21 @@ oc create -f deploy/operator.yaml
 oc create -f deploy/cr.yaml
 ```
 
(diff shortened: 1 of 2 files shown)


ingvagabund:
@mdame PTAL

damemi:
/lgtm cancel
Something I noticed, the CR shouldn't be created until after at least the namespace. So Could you rename the cr to just `02_kube-descheduler-operator.cr.yaml`. This will let you deploy the whole operator with just `oc create -f manifests/.`

damemi:
Picked into https://github.com/openshift/cluster-kube-descheduler-operator/pull/66, so I'm going to close this so we don't need to open another BZ
/close

PR #69, opened 2020-02-04
Bug 1798497: Update deployment generation statuses

damemi:
#66 merged without actually checking any valid value against the expected generation of the deployment. This implements the necessary interfaces to use library-go helpers to update the operator status with the generation information, and uses the information to check against when updating the deployment.


damemi:
/hold
This does update the status generation, but it currently hotloops. Have to figure out where else I need to compare before doing an update

damemi:
/hold cancel
figured it out, need to do some dirty checking to see if we should be forcing a new deployment in the call to `applyDeployment`

ingvagabund:
Don't see anything suspicious

\lgtm

PR #75, opened 2020-02-14
bug 1803139: Add appregistry delivery label to Dockerfile.rhel7

thiagoalessio:
ART relies on the presence of `com.redhat.delivery.appregistry` label to identify it as an OLM operator.


ingvagabund:
/retitle bug 1803139: Add appregistry delivery label to Dockerfile.rhel7

Issue #83, opened 2020-02-25
Support proper strategy names


damemi:
Currently we have this shorthand approach to descheduler strategies that maps a new name to the actual upstream name:

| Operator param | Descheduler strategy |
| ---- | ---- |
| `duplicates` | `RemoveDuplicates` |
| `interpodantiaffinity` | `RemovePodsViolatingInterPodAntiAffinity` |
| `lownodeutilization` | `LowNodeUtilization` |
| `nodeaffinity` | `RemovePodsViolatingNodeAffinity` |
| `nodetaints` | `RemovePodsViolatingNodeTaints` |

This seems confusing and adds something to translate when trying to configure the operator. These are handled in a simple [switch statement](https://github.com/openshift/cluster-kube-descheduler-operator/blob/76aef61bc5399d78b98bcd805b66e78a966621ce/pkg/operator/target_config_reconciler.go#L148) and it should be relatively easy to add support for the real upstream strategy names and make those our primary. We can silently support these shorthands for backward compatibility and phase out eventually

PR #95, opened 2020-03-09
Add support for actual descheduler strategy names in operator config

damemi:
Fixes https://github.com/openshift/cluster-kube-descheduler-operator/issues/83

This will add support to enable descheduler strategies by their actual documented upstream names, which should alleviate confusion and necessary documentation

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -150,9 +150,9 @@ func generateConfigMapString(requestedStrategies []deschedulerv1beta1.Strategy) 
 	// There is no need to do validation here. By the time, we reach here, validation would have already happened.
 	for _, strategy := range requestedStrategies {
 		switch strings.ToLower(strategy.Name) {
-		case "duplicates":
+		case "duplicates", "removeduplicates":
 			policy.Strategies["RemoveDuplicates"] = deschedulerapi.DeschedulerStrategy{Enabled: true}
-		case "interpodantiaffinity":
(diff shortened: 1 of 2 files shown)


damemi:
/cc @ingvagabund 
ptal, this shouldn't change any current behavior because it still silently supports the old configurations, but it will make our documentation much cleaner

pkg/operator/target_config_reconciler.go
```diff
@@ -150,11 +150,11 @@ func generateConfigMapString(requestedStrategies []deschedulerv1beta1.Strategy)
...
 	for _, strategy := range requestedStrategies {
 		switch strings.ToLower(strategy.Name) {
-		case "duplicates":
+		case "duplicates", "RemoveDuplicates":
 			policy.Strategies["RemoveDuplicates"] = deschedulerapi.DeschedulerStrategy{Enabled: true}
-		case "interpodantiaffinity":
+		case "interpodantiaffinity", "RemovePodsViolatingInterPodAntiAffinity":
 			policy.Strategies["RemovePodsViolatingInterPodAntiAffinity"] = deschedulerapi.DeschedulerStrategy{Enabled: true}
-		case "lownodeutilization":
+		case "lownodeutilization", "LowNodeUtilization":
```
soltysh: Why do you need this if you have `ToLower` at the top?

damemi: updated, thanks

PR #100, opened 2020-03-24
Update list of valid strategy names

damemi:


damemi:
Tested this locally and full strategy names are accepted now

PR #103, opened 2020-03-26
Provide manifests for deploying the operator without OLM (for development only)

ingvagabund:
Until we are able to generate all the manifests from the bundle.


ingvagabund:
> Some notes: the CRD should have a prefix that creates it before the CR, like 0000

Does oc respect that?

PR #104, opened 2020-03-27
[release-4.4] Bug 1817318: Fix bug in strategy name validation

damemi:
This backports a critical fix from https://github.com/openshift/cluster-kube-descheduler-operator/pull/100, which allows the documented upstream strategy names to be set in the operator.

It also includes the update from https://github.com/openshift/cluster-kube-descheduler-operator/pull/102, which is a minor change but switches the default operand image to `v0.10.0`, which also includes some critical bugfixes.


damemi:
/assign @ingvagabund 
please lgtm

PR #108, opened 2020-05-06
Bug 1833329: Add support for RemovePodsHavingTooManyRestarts strategy

damemi:
This adds the `RemovePodsHavingTooManyRestarts` strategy, added upstream in https://github.com/kubernetes-sigs/descheduler/pull/254 and synced to our fork in https://github.com/openshift/descheduler/pull/29.

This also updates the sample CR in the readme to show how to configure this strategy, and I made the change to now convert strategy parameter names to lowercase in our logic, allowing CamelCase params that are imo more readable.


pkg/operator/target_config_reconciler.go
```diff
@@ -213,6 +214,29 @@ func generateConfigMapString(requestedStrategies []deschedulerv1beta1.Strategy)
 			}
 		case "nodetaints", "removepodsviolatingnodetaints":
 			policy.Strategies["RemovePodsViolatingNodeTaints"] = deschedulerapi.DeschedulerStrategy{Enabled: true}
+		case "removepodshavingtoomanyrestarts":
+			podsHavingTooManyRestarts := deschedulerapi.PodsHavingTooManyRestarts{}
```
ingvagabund, line 218: Also, let's wait until https://github.com/kubernetes-sigs/descheduler/pull/276 gets merged.

pkg/operator/target_config_reconciler.go
```diff
@@ -177,7 +178,7 @@ func generateConfigMapString(requestedStrategies []deschedulerv1beta1.Strategy)
 				if err != nil {
 					return "", err
 				}
-				switch param.Name {
+				switch strings.ToLower(param.Name) {
```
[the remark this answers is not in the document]
damemi: I don't think that's really necessary, it's just one line (except for the Readme change) and functionally makes no difference to the existing behavior. But if it's easier for review I can if you'd like

pkg/operator/target_config_reconciler.go
damemi: Agreed

pkg/operator/target_config_reconciler.go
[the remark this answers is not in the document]
ingvagabund: There are other changes in README.md going from lower case to upper case. It's no big deal. It just makes clearer which change is adding the new strategy and which change is allowing the params names to be mix of lower and upper cases.

pkg/operator/target_config_reconciler.go
damemi: @ingvagabund the upstream PR has merged, so I updated this as well. Should be all good for a final pass

damemi:
/retitle Bug 1833329: Add support for RemovePodsHavingTooManyRestarts strategy

PR #114, opened 2020-07-13
Bug 1836961: bump(descheduler): Include api changes for StrategyParams

damemi:
Specifically looking for the change from https://github.com/openshift/descheduler/commit/5a81a0661b9fd7226dada62f4a72f955b7a60d71

These are in the actual descheduler image, but not being applied when the operator parses/generates a strategy config because the operator is still vendoring an old version of the api. This bumps that api


damemi:
Confirmed this manually in a cluster:
```
$ oc get -o yaml cm/cluster
apiVersion: v1
data:
  policy.yaml: |
    strategies:
      LowNodeUtilization:
        enabled: true
        params:
          nodeResourceUtilizationThresholds:
            numberOfNodes: 3
            targetThresholds:
              cpu: 10
              memory: 20
              pods: 30
            thresholds:
              cpu: 10
              memory: 20
              pods: 30
      RemoveDuplicates:
        enabled: true
kind: ConfigMap
```

PR #132, opened 2020-09-15
bug 1879139: Be more verbose when the operator configuration is missing

ingvagabund:
When the operator configuration is deployed in different namespace than `openshift-kube-descheduler-operator`,
the operator is reporting:

```
E0825 23:53:11.641464       1 target_config_reconciler.go:307] key failed with : kubedeschedulers.operator.openshift.io "cluster" not found
```

The log does not say anything about the namespace.
Explicitly mentioning the expected namespace may give user more hints on what's wrong.


damemi (approved):
/lgtm

PR #134, opened 2020-09-16
Bug 1879527: Return an error if both include and exclude namespaces are set

damemi:
This change also adds parameter validation to each strategy with a big set
of all possible parameters.


damemi:
/hold
for more discussion on if this will be necessary

damemi (commented):
/close
See related BZ, with the plans to switch to the upstream descheduler policy we shouldn't spend effort getting all of this validation working

PR #136, opened 2020-09-17
bug 1874374: icon: use png instead of svg

ingvagabund:
SVG image format is not properly supported by current browsers.
Thus, the icon is not displayed at all.


soltysh (approved):
/lgtm

PR #138, opened 2020-09-18
[wip] Switch to openshift api/client-go libraries to host Descheduler types

damemi:
This depends on https://github.com/openshift/api/pull/740 and https://github.com/openshift/client-go/pull/160

Removes the old API types and switches to using API/Clients from openshift/api and openshift/client-go


damemi (commented):
/close
We will be keeping this within our repo, see https://github.com/openshift/cluster-kube-descheduler-operator/pull/148

PR #141, opened 2020-10-02
bug 1884601: Drop kube-system role binding to extension-apiserver-authentication-reader

ingvagabund:


soltysh (approved):
/lgtm

deploy/02_clusterrole.yaml
```diff
@@ -28,14 +28,18 @@ rules:
...
   - apps
   resources:
   - deployments
   - replicasets
   verbs:
   - "*"
-- apiGroups: ["batch", "extensions"]
-  resources: ["jobs", "cronjobs"]
-  verbs:
-  - "*"
```
sttts, line 41: this looks like an indepenent change

soltysh: This is cleaning old roles, we're not using. Unlike upstream which relies on cronjob to run descheduler, we have an operator that handles that for us.

damemi: Technically the upstream can also run the descheduler as a control-loop but recommends running it as a cronjob. We chose to deploy it as a control-loop in the operator since cronjobs aren't yet GA (though this operator did deploy it as a cronjob previously, before we recently refactored it).

Once cronjobs are GA, I think it would be good to add that option back into our operator. But obviously we don't need this role until then

ingvagabund: Anything under `deploy` is just for development. Not going into production.

sttts: bad style to mix cleanup and the things described in the commit.

PR #147, opened 2020-10-27
bug 1887740: Extend olm.skipRange check to properly compare pre-release versions

ingvagabund:
When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version so 4.6.0-202010061132.p0 < 4.6.0.
Because of this, the solver is confused by this and adds the this version into `replaces` field so we end up with the situation that this version is replacing itself.
As a result, the CSV is stuck in Pending state.


soltysh (approved):
/lgtm

PR #148, opened 2020-10-29
Switch to accepting upstream Policy

### LINKED ISSUE #119: Switch to actual upstream Descheduler policy
We currently have our own config API in the operator that differs from [the upstream Descheduler API](https://github.com/kubernetes-sigs/descheduler/blob/release-1.18/pkg/api/v1alpha1/types.go). For example, the operator needs to be configured with a `strategies`  field like:
```
apiVersion: operator.openshift.io/v1beta1
kind: KubeDescheduler
metadata:
  name: config
  namespace: openshift-kube-descheduler-operator
spec:
  strategies:
    - name: "RemoveDuplicates"
    - name: "RemovePodsHavingTooManyRestarts"
      params:
       - name: "PodRestartThreshold"
         value: "10"
       - name: "IncludingInitContainers"
         value: "false"
```

which just gets internally translated into:

```
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  "RemoveDuplicates":
     enabled: true
  "RemovePodsHavingTooManyRestarts":
     enabled: true
     params:
       podsHavingTooManyRestarts:
         podRestartThreshold: 10
         includingInitContainers: false
```

So, while our own API is slightly simpler [in its definition](https://github.com/openshift/cluster-kube-descheduler-operator/blob/9ec846ab74fd8b26dc0b4ae13eb73698d8a216f5/pkg/apis/descheduler/v1beta1/types_descheduler.go), in practice it [must be manually converted](https://github.com/openshift/cluster-kube-descheduler-operator/blob/9ec846ab74fd8b26dc0b4ae13eb73698d8a216f5/pkg/operator/target_config_reconciler.go#L163-L249) which adds complexity to the codebase. It also means that we need to constantly update our own operator code to support new strategies and parameters as they are added upstream meaning double the work for us to add a new feature upstream.

In addition, it is also confusing to users that the config API is different when using our operator vs running the descheduler on their own which could inhibit adoption of the operator. It would be much simpler to simply have to point to upstream docs for configuring the descheduler.

This is why I propose a field called `policy` in the operator spec, which would point to a configmap containing an actual descheduler policy (along with an optional field `namespace`, which defaults to `openshift-config`). This would match the design of the [scheduler operator](https://github.com/openshift/cluster-kube-scheduler-operator) which has a Policy field that points to a configmap with a regular scheduler policy (see [the OpenShift docs on how to deploy the scheduler operator with a custom policy](https://docs.openshift.com/container-platform/4.1/nodes/scheduling/nodes-scheduler-default.html). This would be the exact same design).

I think we will still need to support the current `v1beta1` config API until it can be fully deprecated, but this shift will save us effort and reduce potential failure points.

**Update**: I've opened these PRs to begin the work required for this:
* https://github.com/openshift/api/pull/740
* https://github.com/openshift/client-go/pull/160
* https://github.com/openshift/cluster-kube-descheduler-operator/pull/138

damemi:
Fixes #119 

This deprecates the current "strategies" API we define ourselves in the operator in favor of simply accepting an upstream Descheduler policy as a configmap.

As described in the linked issue, this will simplify documentation and stabilize the operator codebase in preparation for bringing the operator to GA.


soltysh (approved):
/lgtm
/hold 
if you want to still add something here
@damemi we still want to GA the api, right? I'm assuming that in v1 API there will only the new bits, right?

damemi:
This actually needs some work, and I think it needs to have an actual configobserver pattern to watch for changes to the policy configmap, then update the operator's `ObservedConfig` with those changes in order to trigger the new deployment. Running this as it is, I get a lot of weird hotlooping right now when I'm just observing configmap+crd in the target config reconciler, which points to what I'm talking about (configmap change causes an operator status/generation update, which then triggers another operator status/generation update)

damemi (commented):
/hold cancel
Added a configobserver approach and in some manual testing it works for me. This probably needs some cleanup but ready for review

pkg/operator/target_config_reconciler.go
```diff
@@ -99,8 +98,14 @@ func (c TargetConfigReconciler) sync() error {
 		return fmt.Errorf("descheduler should have an interval set")
 	}
 
-	if err := validateStrategies(descheduler.Spec.Strategies); err != nil {
-		return err
+	// TODO(@damemi): Remove this check and related functions when Strategies is removed
+	if len(descheduler.Spec.Strategies) > 0 {
```
soltysh, line 102: I think the order should be new field first, with fallback to deprecated field, no?

soltysh (approved):
/lgtm
We can fix the question in a followup.

pkg/operator/target_config_reconciler.go
damemi: Yeah, that's a good point. Will fix in a follow up PR

PR #149, opened 2020-11-02
Update CRD to apiextensions/v1

damemi:
This updates the operator CRD to `apiextensions.k8s.io/v1`, and adds makefile targets to generate the CRD and clients


damemi:
@sttts or @ingvagabund, could you help me debug why the updated CRD isn't recognized by the generated client?

Originally I added this as part of https://github.com/openshift/cluster-kube-descheduler-operator/pull/148, but the controllers all keep failing with this error:

```
E1102 20:08:30.817911       1 target_config_reconciler.go:478] key failed with : kubedeschedulers.operator.openshift.io "cluster" not found
```

To narrow it down, I tried breaking this CRD update out and that seems to be what's causing the failure (I see the same error in an operator deployed with only these changes).

I don't get many changes from the code generator (see the commit where I update the makefile to call `generate-groups.sh`), but I did manually change the CRD to `v1` before calling our schemapatch generator. Any ideas?

damemi (commented):
/close
I had `subresources: status` in the wrong block, doh... change ported to the original PR

PR #153, opened 2020-11-22
Bump 4.7 kubernetes 1.20.0 beta.2

ingvagabund:
SSIA


pkg/operator/target_config_reconciler.go
```diff
@@ -376,7 +376,9 @@ func generateConfigMapString(requestedStrategies []deschedulerv1beta1.Strategy)
 						return "", err
 					}
 					val := uint(value)
-					lifetimeSeconds = &val
+					podLifeTime.MaxPodLifeTimeSeconds = &val
+				case "podstatusphases":
```
ingvagabund, line 380: @damemi A bit hesitant about allowing to set `podStatusPhases` as well. Two options:
1. podlifetime strategy is not a new strategy -> allow to set all fields of the strategy
2. podlifetime has a new field = new feature -> use policy file for anything new (forces consumers to migrate to the policy file)

I am fine with either. Option 2 makes the PR simpler. WDYT?

soltysh (approved):
/lgtm

ingvagabund:
/hold
waiting for @damemi 's review

pkg/operator/target_config_reconciler.go
damemi: I prefer option 2, and would rather focus on whatever our definition of new policy is for new features. But since you already added it, I don't see the harm here.

damemi (commented):
/lgtm

Issue #158, opened 2020-12-03
Descheduler should parse IMAGE env var for development


damemi:
For QA and development, there is a need to be able to set a custom descheduler image in the operator (often to the latest ART build in order to verify bug fixes). This used to be the `image` field in the operator spec, but because we are removing that from the supported user config this should be added as an undocumented/unsupported flag.

For example, the kube scheduler operator (and others) get the [env var in the operator deployment](https://github.com/openshift/cluster-kube-scheduler-operator/blob/c4c246020c43661ee271e7626488977890b9e847/pkg/operator/starter.go#L94-L105) then sub that in with the [config reconciler](https://github.com/openshift/cluster-kube-scheduler-operator/blob/c4c246020c43661ee271e7626488977890b9e847/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L291-L312)

PR #159, opened 2020-12-04
Add support for IMAGE env var

damemi:
Fixes https://github.com/openshift/cluster-kube-descheduler-operator/issues/158


ingvagabund:
/lgtm
/hold

Looks good in overall. Feel free to unhold it.

damemi:
Sorry, forgot to actually update the bindata deployment to have the new env var.

I confirmed with ART that the image-references file in our manifests dir is what's responsible for replacing that value with the managed image hash. So this should be all good now (but will need another lgtm). I also tested locally

/hold cancel

PR #161, opened 2020-12-08
bug 1903188: Set --logging-format to text, empty value generates an error message

ingvagabund:
"failed to validate server configuration" err="unsupported log format: "


damemi (approved):
/approve
/lgtm

PR #177, opened 2021-03-26
Bug 1943224: Fix CSV olmSkipRange pattern for development

damemi:


damemi (commented):
does this need to be backported to the release-4.7 branch too?

ingvagabund:
Ain't gonna fly: https://github.com/openshift/cluster-kube-descheduler-operator/commit/bb8c45f9843d9650559402a1e5a11d4f9e5e37de

damemi:
> When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version so 4.6.0-202010061132.p0 < 4.6.0.

@ingvagabund so, we need the `-0` to be compatible with pre-release versions, but it looks like it breaks on z-stream releases? Do you know how [cluster-logging-operator](https://github.com/openshift/cluster-logging-operator/blob/release-4.5/manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml) is getting away with it?

Maybe we just need the `-0` to support pre-release for our latest (4.8) version, and remember to go back and update once it is released.

damemi:
@ingvagabund could you refresh my memory, what was the outcome of when we talked about this?

ingvagabund:
/lgtm
is the outcome.

PR #178, opened 2021-03-30
Add v1 API

damemi:


damemi (commented):
/cc @soltysh @ingvagabund 
I think I've added the internal API correctly (but running the generate script locally will depend on https://github.com/kubernetes/kubernetes/pull/100666)

Next step is adding the code to actually consume/convert whatever version is passed as a CRD. This will let us bridge existing stored v1beta1 versions to the new v1 version without the extra fields

damemi:
At this point it runs properly for me, but the CRD schema gen needs work to patch it. I just generated this with `controller-gen` manually from the latest upstream branch (rather than our makefile targets, which is fine since we don't need to just patch the openapi schema)

pkg/operator/target_config_reconciler.go
```diff
@@ -3,6 +3,7 @@ package operator
 import (
 	"context"
 	"fmt"
+	deschedulerv1 "github.com/openshift/cluster-kube-descheduler-operator/pkg/apis/descheduler/v1"
```
ingvagabund, line 6: nit: move this next to other `github.com/openshift/cluster-kube-descheduler-operator` imports

pkg/apis/descheduler/v1/doc.go
```diff
@@ -4,5 +4,4 @@
 // +groupName=operator.openshift.io
 // +groupGoName=Kubedeschedulers
 // +kubebuilder:validation:Optional
-// +k8s:conversion-gen=github.com/openshift/cluster-kube-descheduler-operator/pkg/apis/descheduler
```
ingvagabund: This was added in the first commit. Can you remove it there instead?

pkg/apis/descheduler/v1beta1/doc.go
```diff
@@ -4,5 +4,4 @@
 // +groupName=operator.openshift.io
 // +groupGoName=Kubedeschedulers
 // +kubebuilder:validation:Optional
-// +k8s:conversion-gen=github.com/openshift/cluster-kube-descheduler-operator/pkg/apis/descheduler
```
ingvagabund: This was added in the first commit. Can you remove it there instead?

ingvagabund (changes requested):
Just nits

damemi:
Thanks @ingvagabund, I reorganized some of the commits and removed the Makefile changes since those are broken.

So right now, this can't generate a new CRD if there are any changes. I don't think that is technically a blocker, since we have the ability to manually generate any CRD changes should they come up. But it is definitely an issue that could warrant its own bugzilla. @soltysh what do you think about this?

soltysh (approved):
/lgtm

PR #184, opened 2021-04-12
Bug 1943224: Update olmSkipRange in 4.7 and 4.8 to remove dev pattern

damemi:
Follow up to https://github.com/openshift/cluster-kube-descheduler-operator/pull/177, this removes the dev pattern from the lower version numbers in the CSVs

### DIFF
--- a/manifests/4.7/cluster-kube-descheduler-operator.v4.7.0.clusterserviceversion.yaml
+++ b/manifests/4.7/cluster-kube-descheduler-operator.v4.7.0.clusterserviceversion.yaml
@@ -24,7 +24,7 @@ metadata:
     certified: "false"
     containerImage: quay.io/openshift/origin-cluster-kube-descheduler-operator:4.7
     createdAt: 2020/04/18
-    olm.skipRange: ">=4.3.0-0 < 4.7.0"
+    olm.skipRange: ">=4.5.0 < 4.7.0"
     description: An operator to run descheduler in Openshift cluster.
     repository: https://github.com/openshift/cluster-kube-descheduler-operator
     support: Red Hat, Inc.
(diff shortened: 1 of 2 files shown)


damemi (commented):
does this change make sense too? @kasturinarra pointed it out in the linked BZ that the original PR wasn't completely consistent with other operators

kasturinarra:
@damemi looks good to me, but one question when looking at all other operator cvs.yaml files for 4.7 the skiprange defined there is >=4.5.0 <4.7.0, any reason we have >=4.3.0 <4.7.0 ?

ingvagabund:
Wondering how do we actually verify this is the "right" and "final" form of the condition.

damemi:
> Wondering how do we actually verify this is the "right" and "final" form of the condition.

Yeah I am not sure what is the cause for the differences. Maybe those operators have only existed since 4.5, or maybe we should update our supported versions to only be X-3? I don't know if a 4.8 descheduler operator should be guaranteed to run as far back as 4.3

ingvagabund:
So we need to match https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/art.yaml#L10-L11:
```
    - search: 'olm.skipRange: ">=4.6.0 <{MAJOR}.{MINOR}.0"'
      replace: 'olm.skipRange: ">=4.6.0 <{FULL_VER}"'
```

Which is met for 4.8 once this PR merges.

ingvagabund:
>  or maybe we should update our supported versions to only be X-3?

Right. In 4.8, it should be 4.6, 4.7, 4.8. So change in https://github.com/openshift/cluster-kube-descheduler-operator/pull/184/files#diff-b9083740e2642970ffb668b95996958307c31b5b728f79fa9ccb585f8167ef16L27 is actually irrelevant. Though, we might bump 4.3 to 4.5.

kasturinarra:
> > or maybe we should update our supported versions to only be X-3?
> 
> Right. In 4.8, it should be 4.6, 4.7, 4.8. So change in https://github.com/openshift/cluster-kube-descheduler-operator/pull/184/files#diff-b9083740e2642970ffb668b95996958307c31b5b728f79fa9ccb585f8167ef16L27 is actually irrelevant. Though, we might bump 4.3 to 4.5.

@ingvagabund do you plan to do the bump in another patch ?

ingvagabund:
Prefer to do it in this PR.

damemi:
Updated that 4.3 to 4.5

PR #185, opened 2021-04-14
bug 1948640: CRD: add missing subresources.status so the CRD's status can be updated

ingvagabund:
To allow CRD's status to be updated


damemi (commented):
ah, thanks @ingvagabund. This might have gotten dropped from the CRD when I regenerated the whole CRD for `v1`
/lgtm

PR #186, opened 2021-04-15
bug 1949972: servicemonitor: replace exported_namespace with pod_namespace

ingvagabund:
The namespace label is reserved label (see
https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#alertmanagerendpoints).
So the namespace label (indicating from which namespace a pod got
evicted) is automatically relabeled into exported_namespace.
To avoid confusion of what the label means, relabel it into
pod_namespace.

Metrics entry before:
```
descheduler_pods_evicted{endpoint="https", exported_namespace="test1", instance="10.129.3.136:10258", job="metrics", namespace="openshift-kube-descheduler-operator", pod="cluster-85df694595-blktt", result="success", service="metrics", strategy="RemoveDuplicatePods"}
```

Metrics entry after:
```
descheduler_pods_evicted{endpoint="https", exported_namespace="test1", instance="10.129.3.136:10258", job="metrics", namespace="openshift-kube-descheduler-operator", pod="cluster-85df694595-blktt", pod_namespace="test1", result="success", service="metrics", strategy="RemoveDuplicatePods"}
```

### DIFF
--- a/bindata/v4.1.0/kube-descheduler/servicemonitor.yaml
+++ b/bindata/v4.1.0/kube-descheduler/servicemonitor.yaml
@@ -11,6 +11,11 @@ spec:
   endpoints:
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
     interval: 30s
+    metricRelabelings:
+    - action: replace
+      sourceLabels:
+      - exported_namespace
+      targetLabel: pod_namespace
     path: /metrics
     port: https
     scheme: https
--- a/pkg/operator/v410_00_assets/bindata.go
+++ b/pkg/operator/v410_00_assets/bindata.go
@@ -275,6 +275,11 @@ spec:
   endpoints:
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
     interval: 30s
+    metricRelabelings:
+    - action: replace
+      sourceLabels:
+      - exported_namespace
+      targetLabel: pod_namespace
     path: /metrics
     port: https
     scheme: https


damemi (commented):
/lgtm

kasturinarra:
@ingvagabund but one question would not it be confusing because pod=<> shows descheduler cluster pod.

ingvagabund:
> @ingvagabund but one question would not it be confusing because pod=<> shows descheduler cluster pod.

This is not particular about descheduler metrics but about metrics in general. Each metric is injected with `namespace` and `pod` (default behavior) which indicates source of a metric. This is more about learning how metrics data are populated. We can only alter/add labels so a metric is easier to read. Though at the end a user has to sometimes learn more to interpret the metrics properly.

soltysh (approved):
/lgtm

damemi:
We could add another label to indicate which pod was evicted too right?

kasturinarra:
@ingvagabund one more question, the title says replace exported_ namespace with pod_namespace but when i look at the output you showed i see both of them to be present. Is that the same we expected ?

soltysh:
> We could add another label to indicate which pod was evicted too right?

The pod is there already: `pod="cluster-85df694595-blktt"` from the above.

kasturinarra:
> > We could add another label to indicate which pod was evicted too right?
> 
> The pod is there already: `pod="cluster-85df694595-blktt"` from the above.

It corresponds to the descheduler operand pod and not the one which is actually evicted.

ingvagabund:
> We could add another label to indicate which pod was evicted too right?

The idea was to be strategy/namespace oriented, not pod oriented

ingvagabund:
> @ingvagabund one more question, the title says replace exported_ namespace with pod_namespace but when i look at the output you showed i see both of them to be present. Is that the same we expected ?

You, I did not manage to have the drop rule work properly :-|

kasturinarra:
> > @ingvagabund one more question, the title says replace exported_ namespace with pod_namespace but when i look at the output you showed i see both of them to be present. Is that the same we expected ?
> 
> You, I did not manage to have the drop rule work properly :-|

hmm, so are we going to revert this change then ?

ingvagabund:
> hmm, so are we going to revert this change then ?

Why?

kasturinarra:
> > hmm, so are we going to revert this change then ?
> 
> Why? Keeping the old label does not hurt.

hmm, okay, let me verify one changes are in , based on that i can comment more, thanks !!

PR #189, opened 2021-04-27
bug 1953707: Have olm.skipRange in csv correspond to search pattern in art.yaml

ingvagabund:
olm.skipRange string has to match with a search pattern in art.yaml to
have it properly replaced with the right version


damemi (commented):
/lgtm

PR #191, opened 2021-04-27
Bug 1955595: Add DevPreviewLongLifecycle profile

damemi:
Re: https://issues.redhat.com/browse/WRKLDS-286
This adds a new Beta profile `BetaLongLifecycle` which is similar to `LifecycleAndUtilization` but which does not evict running pods that exceed the latter's 24-hour lifetime limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,9 @@ The following profiles are currently provided:
 * [`AffinityAndTaints`](#AffinityAndTaints)
 * [`TopologyAndDuplicates`](#TopologyAndDuplicates)
 * [`LifecycleAndUtilization`](#LifecycleAndUtilization)
+  
+Along with the following profiles, which are in development and may change:
+* [`DevPreviewLongLifecycle`](#DevPreviewLongLifecycle)
 
 Each of these enables cluster-wide descheduling (excluding openshift and kube-system namespaces) based on certain goals.
 
@@ -100,6 +103,10 @@ This profile enables the [`LowNodeUtilizaition`](https://github.com/kubernetes-s
 [`RemovePodsHavingTooManyRestarts`](https://github.com/kubernetes-sigs/descheduler/#removepodshavingtoomanyrestarts) and 
 [`PodLifeTime`](https://github.com/kubernetes-sigs/descheduler/#podlifetime) strategies. In the future, more configuration
 may be made available through the operator for these strategies based on user feedback.
+
+### DevPreviewLongLifecycle
+This profile provides cluster resource balancing similar to [LifecycleAndUtilization](#LifecycleAndUtilization) for longer-running 
+clusters. It does not evict pods based on the 24 hour lifetime used by LifecycleAndUtilization.
 
 ## How does the descheduler operator work?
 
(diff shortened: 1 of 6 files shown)


damemi (commented):
/cc @ingvagabund @soltysh 
I think this is all the work that's needed to add the new profile. PTAL and we can discuss any changes or anything I missed

README.md
```diff
@@ -74,6 +74,9 @@ The following profiles are currently provided:
 * [`AffinityAndTaints`](#AffinityAndTaints)
 * [`TopologyAndDuplicates`](#TopologyAndDuplicates)
 * [`LifecycleAndUtilization`](#LifecycleAndUtilization)
+  
+Along with the following profiles, which are in Beta and may change:
+* [`BetaLongLifecycle`](#BetaLongLifecycle)
```
ingvagabund, line 79: Why Beta, why not Alpha?

soltysh: Since this is externally visible let's align that with our terminology and use `DevPreview` or `TechPreview`, the former is equivalent of alpha, the latter - beta.

soltysh (commented):
I agree with Jan, let's change the name to something familiar to OCP, otherwise this lgtm.

manifests/4.8/kube-descheduler-operator.crd.yaml
```diff
@@ -10,8 +12,6 @@ spec:
     plural: kubedeschedulers
     singular: kubedescheduler
   scope: Namespaced
-  subresources:
-    status: {}
```
damemi, line 14: This top-level `subresources` field was removed, and replaced with one in each version below when I ran the generator... @ingvagabund will this work or is it the same problem you fixed in https://github.com/openshift/cluster-kube-descheduler-operator/pull/185?

damemi:
Updated with the feedback. I also added a check for profile conflicts (to ensure that the new profile isn't enabled alongside the old one, since we have no defined way to resolve that)

manifests/4.8/kube-descheduler-operator.crd.yaml
ingvagabund: Yeah, it should work.

pkg/operator/target_config_reconciler.go
```diff
@@ -242,6 +248,21 @@ func (c *TargetConfigReconciler) manageConfigMap(descheduler *deschedulerv1.Kube
 	return resourceapply.ApplyConfigMap(c.kubeClient.CoreV1(), c.eventRecorder, required)
 }
 
+// checkProfileConflicts ensures that multiple profiles aren't redeclared
+// it also checks for various inter-profile conflicts (profiles which should not be enabled simultaneously)
+func checkProfileConflicts(profiles sets.String, profileName deschedulerv1.DeschedulerProfile) error {
```
ingvagabund, line 253: You wanna write a unit test for it? :P

manifests/4.8/kube-descheduler-operator.crd.yaml
soltysh: Yeah, v1beta1 had it here, v1 has it per specific version.

pkg/operator/target_config_reconciler.go
soltysh: Would be nice, but can be done in a followup.

soltysh (approved):
/lgtm

PR #196, opened 2021-05-26
Bug 1964270: Shorten CSV name to fit within kube label limit of 63char

damemi:


damemi (commented):
/cc @kasturinarra @ingvagabund 
@kevinrizza mentioned offline that this should be okay to do, but we should get someone from OLM to double check

ingvagabund:
/close
In favor of https://issues.redhat.com/browse/ART-3002

PR #198, opened 2021-06-08
Bug 1989724: Add Profiles to evict localStorage pods/ignore PVC pods

damemi:
Adds profiles which will enable these options in the descheduler policy
ref: https://issues.redhat.com/browse/WRKLDS-304

### DIFF
--- a/bindata/assets/profiles/EvictPodsWithLocalStorage.yaml
+++ b/bindata/assets/profiles/EvictPodsWithLocalStorage.yaml
@@ -1,0 +1,3 @@
+apiVersion: "descheduler/v1alpha1"
+kind: "DeschedulerPolicy"
+evictLocalStoragePods: true
--- a/manifests/4.9/kube-descheduler-operator.crd.yaml
+++ b/manifests/4.9/kube-descheduler-operator.crd.yaml
@@ -82,6 +82,8 @@ spec:
                       - LifecycleAndUtilization
                       - DevPreviewLongLifecycle
                       - SoftTopologyAndDuplicates
+                      - EvictPodsWithLocalStorage
+                      - DoNotEvictPodsWithPVC
                 unsupportedConfigOverrides:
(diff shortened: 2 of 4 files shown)


damemi (commented):

manifests/4.9/kube-descheduler-operator.crd.yaml
```diff
@@ -92,6 +92,10 @@ spec:
                   - TopologyAndDuplicates
                   - LifecycleAndUtilization
                   - DevPreviewLongLifecycle
+                  - EvictPodsWithLocalStorage
+                  - DoNotEvictPodsWithPVC
+                  type: string
```
ingvagabund, line 97: duplicated line

manifests/4.9/kube-descheduler-operator.crd.yaml
```diff
@@ -92,6 +92,10 @@ spec:
...
+                type: array
```
ingvagabund, line 98: duplicated line

bindata/assets/profiles/EvictPodsWithLocalStorage.yaml
```diff
@@ -0,0 +1,3 @@
+apiVersion: "descheduler/v1alpha1"
+kind: "DeschedulerPolicy"
+evictLocalStoragePods: true
```
ingvagabund, line 3: From https://github.com/openshift/cluster-kube-descheduler-operator/blob/c9a04369033fb1b4476d6577b60511f6cf25a866/pkg/operator/target_config_reconciler.go#L238:
```
// Merge will fill any empty for value type attributes on the dst struct using corresponding
// src attributes if they themselves are not empty. dst and src must be valid same-type structs
// and dst must be a pointer to struct.
// It won't merge unexported (private) fields and will do recursively any exported field.
func Merge(dst, src interface{}, opts ...func(*Config)) error {
	return merge(dst, src, opts...)
}
```

empty fields (i.e. ignorePvcPods and evictLocalStoragePods set to false) will get set to true if set through the profiles. Looks good.

ingvagabund (approved):

ingvagabund (changes requested):
Removing the duplicated lines is the only nit

PR #199, opened 2021-06-08
Remove v1beta1 API

damemi:


damemi (commented):
/cc @soltysh @ingvagabund 
this should probably be the first change we make for 4.9 in the descheduler operator, since all of the other params and features we're adding will build on it.

soltysh:
@damemi can you add a note in the release note pls https://github.com/openshift/openshift-docs/issues/33497

PR #201, opened 2021-06-16
Bug 1989720: Add PodLifetime parameter

damemi:
x-ref: https://issues.redhat.com/browse/WRKLDS-301


pkg/apis/descheduler/v1/types_descheduler.go
```diff
@@ -34,6 +35,10 @@ type KubeDeschedulerSpec struct {
 	// DeschedulingIntervalSeconds is the number of seconds between descheduler runs
 	// +optional
 	DeschedulingIntervalSeconds *int32 `json:"deschedulingIntervalSeconds,omitempty"`
+
+	// PodLifetime is the length of time after which pods should be evicted
+	// This field should be used with profiles that enable the PodLifetime strategy, such as LifecycleAndUtilization
+	PodLifetime *time.Duration `json:"podLifetime,omitempty"`
```
soltysh, line 41: I'm trying to think through how to best approach those customizations. This proposed approach imo isn't the best.
I'm thinking of something like:
```go
type KubeDeschedulerSpec struct {
    ...
    // ProfileCustomizations allows to customize default parameters of any given profile
    ProfileCustomizations ProfileCustomizations
}

type ProfileCustomizations struct {
    // AffinityAndTaintsCustomizations allows providing custom values for AffinityAndTains profile
    AffinityAndTaintsCustomizations *AffinityAndTainsCutomizations 
   ...
}
```
Basically, we'd have struct for each profile holding values specific for it. We allow multiple profiles being specified so we need to similarly be able to tweak parameters for each of them distinctively. 
What do you think about that kind of approach? 
@ingvagabund opinions?
@deads2k for API related review/thoughts

ingvagabund: Putting customization of all strategies under `ProfileCustomizations` makes it more obvious it's a customization.
+1 for this approach. Assuming all profiles that can be configured alongside have disjoint set of strategies so there's clear mapping.

damemi: I think that may be going to much toward re-writing the upstream API, which we wanted to avoid.

> Assuming all profiles that can be configured alongside have disjoint set of strategies

We already don't have this with `LifecycleAndUtilization` and `DevPreviewLongLifecycle`. I imagine we will probably introduce more overlapping/conflicting profiles in the future, so I think trying to directly map options to profiles could end up with a lot of redundant fields.

Another example is when we add a profile for `HighNodeUtilization`, that profile can't be enabled alongside `LifecycleAndUtilization` (LowNodeUtilization) but it could re-use the same threshold parameters.

So, I agree in that a good compromise would be to put the options under a new field `ProfileCustomizations`, like you suggested, but I do not think they should be another layer further under profile-specific fields

soltysh: Spoke with @damemi on slack, yeah a single `ProfileCustomizations` is the preferred way to go forward.

manifests/4.9/kube-descheduler-operator.crd.yaml
```diff
@@ -11,167 +11,144 @@ spec:
...
-                    profiles for the descheduler it allows multiple profiles to be
-                    enabled at once, which will have cumulative effects on the cluster.
+    - name: v1
+      schema:
+        openAPIV3Schema:
+          description: KubeDescheduler is the Schema for the deschedulers API
+          type: object
+          properties:
+            apiVersion:
+              description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
damemi, line 21: @soltysh this seems to be undoing some of the changes from your updates in https://github.com/openshift/cluster-kube-descheduler-operator/pull/205/commits/53cadec8cf411975c075f8258c4da595b5bf75f6

This was run with just `make update-codegen-crds` (when I could get it to run and not segfault)... how did you make your changes there?

damemi (commented):
/cc @ingvagabund 
ptal

manifests/4.9/kube-descheduler-operator.crd.yaml
ingvagabund: The entire block is indented to the right a bit. Any way to move it to the original column?

ingvagabund: > when I could get it to run and not segfault
Can you more elaborate on how did you get around the segfault?

damemi: I just kept re-running it until it passed. I think there is some sort of race in our version of the generator

pkg/operator/target_config_reconciler.go
```diff
@@ -276,6 +276,14 @@ func (c *TargetConfigReconciler) manageConfigMap(descheduler *deschedulerv1.Kube
 		return nil, true, fmt.Errorf("enabling Descheduler LowNodeUtilization with Scheduler HighNodeUtilization may cause an eviction/scheduling hot loop")
 	}
 
+	// set PodLifetime if non-default
+	if descheduler.Spec.ProfileCustomizations.PodLifetime != nil {
```
ingvagabund, line 280: descheduler.Spec.ProfileCustomizations can be nil as well:
```
if descheduler.Spec.ProfileCustomizations != nil && descheduler.Spec.ProfileCustomizations.PodLifetime != nil {
```

damemi: Good point, updated

PR #207, opened 2021-08-05
Bug 1990603: Fix unable to find scheduler config

damemi:
Fixes error caused by https://github.com/openshift/cluster-kube-descheduler-operator/pull/202, which forgot to start the new informer and add permissions for the `scheduler.config.openshift.io` API group

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"context"
 	"fmt"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


damemi (commented):
/cc @ingvagabund 
I also changed how this scales the deployment, because the other way caused the error message to keep getting logged. All we need to do is set `replicas: 0` rather than calling the entire `manageDeployment` func again.

Tested this manually and it looks like it's working

pkg/operator/target_config_reconciler.go
```diff
@@ -115,26 +116,35 @@ func (c TargetConfigReconciler) sync() error {
...
 				v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
 					Type:   "TargetConfigControllerDegraded",
 					Status: operatorv1.ConditionTrue,
-				}),
-				func(status *operatorv1.OperatorStatus) error {
-					resourcemerge.SetDeploymentGeneration(&status.Generations, deployment)
-					return nil
-				})
+				}))
+			// return nil so the controller doesn't keep logging the error and the queue key is forgotten
```
ingvagabund, line 146: The key needs to be forgotten only when both the operator is in Degraded state and the descheduler deployment is scaled to 0. Otherwise, the descheduler might keep running or a user does not get notified. Unless there's another event that periodically triggers the reconciling loop which is independent of the scheduler config.

damemi: Good point, updated this to just return `err` (if `nil`, great, if not then it will report+retry)

PR #208, opened 2021-08-09
Bug 1989720: Add profile customizations to README

damemi:
Follow up https://github.com/openshift/cluster-kube-descheduler-operator/pull/201 to update readme describing profileCustomizations (and example in sample CR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,9 @@ spec:
   deschedulingIntervalSeconds: 1800
   profiles:
   - AffinityAndTaints
+  - LifecycleAndUtilization
+  profileCustomizations:
+    podLifetime: 5m
 ```
 
 The operator spec provides a `profiles` field, which allows users to set one or more descheduling profiles to enable.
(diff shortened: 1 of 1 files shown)


damemi (commented):
/cc @ingvagabund 
could you please tag? Forgot to add these to the readme

README.md
```diff
@@ -61,6 +61,9 @@ spec:
   deschedulingIntervalSeconds: 1800
   profiles:
   - AffinityAndTaints
+  - LifecycleAndUtilization
+  profileCustomizations:
+    podLifetime: 5m
```
kasturinarra, line 66: @damemi above does not work, it throws us an error which reads as below, so the value should be only integer

spec.profileCustomizations.podLifetime: Invalid value: "string": spec.profileCustomizations.podLifetime in body must be of type integer: "string"

damemi: Opened a new PR to address this: https://github.com/openshift/cluster-kube-descheduler-operator/pull/211

PR #210, opened 2021-08-11
bug 1992478: Set served=false for v1beta1 crd

ingvagabund:
Given 4.9 removes v1beta1 version, the OLM complains with
"CRD removes version v1beta1 that is listed as a stored version on the existing CRD".
Based on https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview
we need to set served=false in v1beta1 version before it can be removed.


damemi (commented):
/approve
/lgtm

PR #211, opened 2021-08-16
Bug 1989720: Update CRD to accept PodLifetime as duration

damemi:
Switch to `metav1.Duration` so that this gets generated as a `string` validator, and add an additional requirement for `format: duration`


damemi (commented):
/cc @soltysh @kasturinarra

soltysh (approved):
/lgtm
/approve

damemi:
@soltysh could you re-tag please?

PR #213, opened 2021-08-20
Bug 1989720: Ensure PodLifetime customization parameter is copied

damemi:
This is another (hopefully last) follow up to get podlifetime customization working. It also adds a test to ensure the value is correctly copied to the policy configmap


damemi (commented):
/cc @ingvagabund 
please take a look

PR #217, opened 2021-09-01
Bug 1997787: Rename DoNotEvictPodsWithPVC->EvictPodsWithPVC

damemi:
In https://github.com/openshift/cluster-kube-descheduler-operator/pull/168 we added default behavior to the operator to ignore pods with PVCs (the upstream default is to evict pods with PVCs, but we had a specific bug to ignore them).

Therefore, adding a profile to ignore pods with PVCs was redundant and unclear in context of the existing default behavior of the operator. Before we release `DoNotEvictPodsWithPVC` in 4.9, this changes it to `EvictPodsWithPVC` which allows PVC pods to be evicted.


damemi (commented):
/cc @ingvagabund @bergerhoffer @kasturinarra 
See my description, but in summary we overlooked the fact that ignoring PVC pods is already the default of the operator (it is not default upstream). So I think pivoting this profile is the best solution

PR #218, opened 2021-09-07
Bug 2000653: Add hypershift to list of default excluded namespaces

damemi:


damemi (commented):
/hold
pending confirmation of any other hypershift-related namespaces that need to be added

PR #215, opened 2021-08-30
Bug 1991938: Update stored version to v1

damemi:
Re: https://github.com/openshift/cluster-kube-descheduler-operator/pull/209

From [slack discussion](https://coreos.slack.com/archives/GK58XC2G2/p1630068955008300) we need to manually update the stored version of the CRD in 4.8 so that upgrades can work (v1beta1 was removed in 4.9)

PR #223, opened 2021-09-24
[release-4.8] Bug 1991938: Fix logic check in migration for 4.8

damemi:
Follow up a logical error in https://github.com/openshift/cluster-kube-descheduler-operator/pull/215 that caused this check to fail with `key failed with : storageversionmigrations.migration.k8s.io "operator-kubedescheduler-storage-version-migration" not found` (due to the fact that `Get()` returns a non-nil object when returning an error)


damemi (commented):
/cc @ingvagabund

damemi:
@soltysh can you override the bugzilla label like we did before? This only affects 4.8 so there will be no dependent bug

soltysh:
This PR is only applied to 4.8 so it's pointless to create 4.9 BZ. I'm manually applying the valid-label on it.

dhellmann:
[patch-manager] :rocket: Approved for z-stream by score: 0.50

picked

PR #224, opened 2021-10-05
[release-4.8] Bug 1991938: Switch to non-protobuf config for migrations client

damemi:
New error in our attempts to fix this bug: `E1004 10:24:41.091323       1 target_config_reconciler.go:425] key failed with : object *v1alpha1.StorageVersionMigration does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message`


damemi (commented):
/cc @soltysh 
to override `invalid-bug` label

soltysh:
Overriding valid-bug label, since this is only for 4.8.

Issue #225, opened 2021-10-13
[Release] Please prepare bundle for OCP 4.10


thegreyd:
Now that we are fully branched for 4.10, please prepare your operator to supply a 4.10 bundle (https://github.com/openshift/cluster-kube-descheduler-operator/tree/release-4.10/manifests), so that 4.10 operator publishing works

PR #226, opened 2021-10-13
Add 4.10 operator bundle manifests

damemi:
Fixes https://github.com/openshift/cluster-kube-descheduler-operator/issues/225


ingvagabund (changes requested):
Also https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/art.yaml and https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/cluster-kube-descheduler-operator.package.yaml needs to be updated. You can also completely remove 4.6 to 4.9 directories.

PR #232, opened 2021-11-25
Several smaller cleanups in descehduler-operator

soltysh:
/assign @atiratree 
/cc @damemi @ingvagabund


ingvagabund:
/lgtm

We used the Dockerfile in the past for testing. Though, we don't use it anymore given it points to 4.4.

PR #235, opened 2022-01-05
Bug 2035986: Add default-container annotation

deejross:
Adds new `default-containter` annotation to deployment.

/assign @soltysh


damemi (commented):
/approve
/lgtm

PR #237, opened 2022-01-18
bug 2041581: Allow to set ciphers through observed config

ingvagabund:
To support cluster wide cryptographic policy as is done in KS, KA and KCM.

Originally reported in https://bugzilla.redhat.com/show_bug.cgi?id=1844989 and resolved as part of https://issues.redhat.com/browse/WRKLDS-241 for KS, KA and KCM

Following the same code changes as performed in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/506.


soltysh (approved):
/lgtm

PR #238, opened 2022-01-18
bug 2041466: Dockerfile: use make build instead of go build

ingvagabund:
`go build` does not take into account env variables populated by the building system. `make build` does by populating GO_LD_FLAGS env (see https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/vendor/github.com/openshift/build-machinery-go/make/lib/golang.mk).


soltysh (approved):
/lgtm

PR #239, opened 2022-01-19
Bug 2042029: add missing permission for leases.coordination.k8s.io

soltysh:
After switching to leases.coordination.k8s.io in dafe1bde7ad5aa7656dd7e087f068147a1d357d7
we need to ensure that there exists necessary permissions for it.

/assign @ingvagabund @ravisantoshgudimetla


soltysh:
Done in https://github.com/openshift/cluster-kube-descheduler-operator/pull/237

PR #246, opened 2022-03-17
bug 1991938: Add missing rbac

ingvagabund:
Add missing RBAC rule for updating kubedeschedulers.operator.openshift.io CRD status
```
E0316 15:26:00.119302       1 target_config_reconciler.go:425] key failed with : customresourcedefinitions.apiextensions.k8s.io "kubedeschedulers.operator.openshift.io" is forbidden: User "system:serviceaccount:openshift-kube-descheduler-operator:openshift-descheduler" cannot update resource "customresourcedefinitions/status" in API group "apiextensions.k8s.io" at the cluster scope
```

### DIFF
--- a/deploy/02_clusterrole.yaml
+++ b/deploy/02_clusterrole.yaml
@@ -56,3 +56,15 @@ rules:
   - replicasets
   verbs:
   - "*"
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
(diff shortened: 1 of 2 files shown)


soltysh:
Manually overriding valid-bug label since this is only targetted at 4.8.
/retest-required

PR #260, opened 2022-05-25
bug 2091033: OSDOCS-3679: Clarifying which mode simulates vs evicts

bergerhoffer:


ingvagabund:
/retitle bug 2091033: OSDOCS-3679: Clarifying which mode simulates vs evicts

PR #272, opened 2022-09-07
Use ApplyDeployment instead of ApplyDeploymentWithForce

knelasevero:
Use ApplyDeployment instead of ApplyDeploymentWithForce 

since in ApplyDeployment it sets an anotation that is the hash of the spec: https://github.com/openshift/library-go/blob/245d1ca16d1514313f3a7ab13ff2c9e3b07b5f43/pkg/operator/resource/resourceapply/apps.go#L95

and ApplyDeployment calls ApplyDeploymentWithForce, which then checks that hash to decide or not to force apply: https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L131

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -3,7 +3,6 @@ package operator
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -31,7 +30,6 @@ import (
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -117,8 +115,11 @@ func (c TargetConfigReconciler) sync() error {
 		return fmt.Errorf("descheduler should have an interval set")
 	}
 
-	forceDeployment := false
-	_, forceDeployment, err = c.manageConfigMap(descheduler)
+	specAnnotations := map[string]string{
+		"kubedeschedulers.operator.openshift.io/cluster": strconv.FormatInt(descheduler.Generation, 10),
+	}
+
+	configMap, forceDeployment, err := c.manageConfigMap(descheduler)
 	if err != nil {
 		// if we returned an error from the configmap AND want to force a deployment
 		// it means we want to scale the deployment to 0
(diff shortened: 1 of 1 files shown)


pkg/generated/clientset/versioned/fake/register.go
```diff
@@ -21,14 +21,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
```
ingvagabund, line 24: Any way to undo this changes?

pkg/generated/clientset/versioned/scheme/register.go
```diff
@@ -21,14 +21,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
```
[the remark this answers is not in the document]
knelasevero: Done :heavy_check_mark:

pkg/operator/target_config_reconciler.go
```diff
@@ -117,35 +115,7 @@ func (c TargetConfigReconciler) sync() error {
 		return fmt.Errorf("descheduler should have an interval set")
 	}
 
-	forceDeployment := false
-	_, forceDeployment, err = c.manageConfigMap(descheduler)
```
ingvagabund, line 121: This needs to be handled separately to `ApplyDeployment` as a new annotation. Since the hash calculated from the deployment itself does not include changes in the external dependent resources. There are two cases to take into account:
1. config map changes: the deployment needs to be annotated with `"configmaps/cluster": configMap.ObjectMeta.ResourceVersion` to trigger additional update of the deployment
2. if err != nil, the deployment needs to be scaled down: the original code needs to be kept as it is. Just, without passing `forceDeployment` into `manageDeployment`. `forceDeployment` will be used only for scaling the deployment to `0`.

ingvagabund (changes requested):

pkg/operator/target_config_reconciler.go
knelasevero: I am annotating all the other dependent resources (services, roles, sa, etc). Let me know if we just want to annotate for the configMap

pkg/operator/target_config_reconciler.go
```diff
@@ -146,38 +147,51 @@ func (c TargetConfigReconciler) sync() error {
 					Status: operatorv1.ConditionTrue,
 				}))
 		}
-		return err
```
ingvagabund, line 149: We still want to return `err` here.

pkg/operator/target_config_reconciler.go
ingvagabund: +1 for annotating the other resources

pkg/operator/target_config_reconciler.go
```diff
@@ -146,38 +147,51 @@ func (c TargetConfigReconciler) sync() error {
...
+	} else {
+		specAnnotations["configmaps/cluster"] = configMap.ObjectMeta.ResourceVersion
```
ingvagabund, line 151: Worth double checking https://github.com/openshift/cluster-openshift-controller-manager-operator/commit/5e87d418fa9e170ee615109b9875b92427d7cd16 whether all objects are non-nil. The same for SSO.

PR #287, opened 2023-02-28
simple podlifetime eviction e2e

### LINKED ISSUE #288: e2e is only checking operator, not operand
e2e is only checking operator, not operand and we need to check both.

So this message is wrong: https://github.com/openshift/cluster-kube-descheduler-operator/blob/3cc94e27dfc7eedbec34cf4cc1f6807ee28d2cc4/test/e2e/operator_test.go#L191

And we need one more check.

knelasevero:
I was trying to come up with a few more interesting tests, but could not get it to work as I intended. 

But I guess that from the operator perspective we can just test this simple podlifetime eviction and it is enough

Also Closes #288


test/e2e/bindata/assets/03_clusterrole.yaml
```diff
@@ -1,7 +1,7 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: descheduler-operator
+  name: openshift-descheduler-operator
```
ingvagabund, line 4: The naming needs to stay identical to what's used in https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/stable/cluster-kube-descheduler-operator.v4.14.0.clusterserviceversion.yaml#L111-L195. So we test what we deliver. At best, the e2e test should read the CSV file and extract the naming and the list of RBAC rules from it. So there's no disconnection. Otherwise, everytime the list of rbac rules in the CSV file gets updated, we need to manually mirror the changes here.

test/e2e/bindata/assets/04_descheduler-cluster-role-binding.yaml
```diff
@@ -1,12 +1,12 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: openshift-descheduler-operator
+  name: openshift-descheduler
```
ingvagabund, line 4: Is this meant to be for the operand or the operator?

knelasevero: That's why this repository is confusing :sweat_smile: Things sometimes are called as simply "descheduler" and sometimes as "descheduler-operator" and it is not clear to which things are actually referring.

test/e2e/bindata/assets/03_clusterrole.yaml
knelasevero: Would be cool, but won't be doing that one in this PR :D

test/e2e/operator_test.go
```diff
@@ -163,8 +165,8 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
-	var deschPod *corev1.Pod
-	// Wait until the descheduler pod is running
+	var deschOpPod *corev1.Pod
+	// Wait until the descheduler pod operator is running
```
ingvagabund, line 183: descheduler operator pod

test/e2e/bindata/assets/02_operandserviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: openshift-descheduler-operand
```
ingvagabund, line 4: You should not need to deploy the operand SA and CR. The operator is responsible for rendering those:
- https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/bindata/assets/kube-descheduler/operandclusterrole.yaml
- https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/bindata/assets/kube-descheduler/operandserviceaccount.yaml
- https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/bindata/assets/kube-descheduler/operandclusterrolebinding.yaml

test/e2e/bindata/assets/05_deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
-          image: # set in e2e
+          image: quay.io/openshift/origin-cluster-kube-descheduler-operator:4.14
```
ingvagabund, line 28: 4.13 :( After the branching we will need to bump to 4.14 again.

test/e2e/bindata/assets/05_deployment.yaml
```diff
@@ -46,6 +46,5 @@ spec:
             - name: OPERATOR_NAME
               value: "descheduler-operator"
             - name: IMAGE
-              value: # set in e2e
+              value: quay.io/openshift/origin-descheduler:4.14
```
ingvagabund, line 49: 4.13 :( After the branching we will need to bump to 4.14 again.

test/e2e/bindata/assets/05_deployment.yaml
ingvagabund, line 28: Does this have any affect?

test/e2e/bindata/assets/05_deployment.yaml
ingvagabund, line 49: Does this have any affect?

test/e2e/operator_test.go
```diff
@@ -186,20 +201,52 @@ func TestMain(m *testing.M) {
...
 	}); err != nil {
-		klog.Errorf("Unable to wait for the SS pod to run")
+		klog.Errorf("Unable to wait for the Descheduler operator pod to run")
 		os.Exit(1)
 	}
+	klog.Infof("Descheduler opearator running in %v", deschOpPod.Name)
 
-	klog.Infof("Descheduler operator running in %v", deschOpPod.Name)
+	var deschPod *corev1.Pod
+	// Wait until the descheduler pod operator is running
```
ingvagabund, line 210: descheduler operator pod

test/e2e/operator_test.go
```diff
@@ -186,20 +201,52 @@ func TestMain(m *testing.M) {
 		}
 		return false, nil
 	}); err != nil {
-		klog.Errorf("Unable to wait for the SS pod to run")
+		klog.Errorf("Unable to wait for the Descheduler operator pod to run")
 		os.Exit(1)
 	}
+	klog.Infof("Descheduler opearator running in %v", deschOpPod.Name)
```
ingvagabund, line 207: opearator -> operator

test/e2e/bindata/assets/05_deployment.yaml
knelasevero: No, I changed this as I was trying applying the folder directly into a cluster to check things running. I will change it back when things are working.

Can we focus on checking why the test pod is not getting evicted?

test/e2e/bindata/assets/02_operandserviceaccount.yaml
knelasevero: Ok

PR #298, opened 2023-05-12
OCPBUGS-13579: WRKLDS-700: bump(k8s) to v0.27.1

knelasevero:
title says it all


ingvagabund:
https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/go.mod#L3 needs to be bumped to 1.20

PR #301, opened 2023-09-12
use `stable` channel in CSV

sosiouxme:
With 4.14, ART no longer forces a stable channel, but instead uses what you have here as the channel list. You probably want it to be `stable`.

reference [slack chat](https://redhat-internal.slack.com/archives/CB95J6R4N/p1694127434005319?thread_ts=1694094694.738739&cid=CB95J6R4N) and you'll want to backport this to 4.14 as well.


ingvagabund:
/retitle OCPBUGS-20312: use stable channel in CSV

PR #346, opened 2024-05-07
wrong changes

dmrost:
Fix wrong path for 4.15


ingvagabund:
Hello @dmrost. This dockerfile is no longer in use. In 4.15 we changed the way to operator images are build. We actually need to delete Dockerfile.metadata. Do you have any use for it?

ingvagabund:
@dmrost the file can stil be used for testing purposes. Though, only for that.

PR #350, opened 2024-05-21
fix: descheduler deployment wrong name

ytinirt:
While we met error from the configmap and force a deployment scale to 0, the name of descheduler deployment is wrong.

error log by `target_config_reconciler.go` is
```
key failed with : deployments.apps "cluster" not found
```


ingvagabund:
@ytinirt nice catch!!! Thank you

PR #384, opened 2024-10-16
WRKLDS-1492: Update k8s dependencies to 1.31.1

ingvagabund:
Generating `ExtractSecondaryScheduler` and `ExtractSecondarySchedulerStatus` by hand


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -45,6 +45,7 @@ func NewConfigObserver(
 
 	c := &ConfigObserver{
 		Controller: configobserver.NewConfigObserver(
+			"cluster-kube-descheduler-operator",
```
ardaguclu, line 48: I see the usages of operand name rather than operator. It may not create any difference but it is worth to mention.

ingvagabund: I wondered about that. Though, on the other hand it's the operator that observes the configs. I wonder whether there's any best practice for this.

ardaguclu: If you are fine with the current value, I can tag this.

ingvagabund: I can see how both names are valid. Let's keep it as it is. Thx for the review Arda :)

PR #459, opened 2025-03-06
OCPBUGS-52341: Add CPUPSIPressureUtilization ActualUtilizationProfile

tiraboschi:
CPU PSI pressure as measured by the Kernel is not able to discriminate self imposed pressure due to CPU throttling on CPU limited containers from pressure caused by resource contention with other neighbors.

See: https://github.com/kubernetes/enhancements/issues/5062

In order to filter out evident false positives, we can introduce a new ActualUtilizationProfile that is consuming CPU PSI pressure value only when the actual average CPU utilization on the node is above a certain threshold (70%), otherwise 0.
This is not a systematic solution but a mitigation.

Fixes: https://issues.redhat.com/browse/OCPBUGS-52341

### DIFF
--- a/pkg/apis/descheduler/v1/types_descheduler.go
+++ b/pkg/apis/descheduler/v1/types_descheduler.go
@@ -129,6 +129,8 @@ const (
 	PrometheusCPUUsageProfile ActualUtilizationProfile = "PrometheusCPUUsage"
 	// PrometheusCPUPSIPressureProfile sets rate(node_pressure_cpu_waiting_seconds_total[1m]) query
 	PrometheusCPUPSIPressureProfile ActualUtilizationProfile = "PrometheusCPUPSIPressure"
+	// PrometheusCPUPSIPressureUtilizationProfile sets a query based on a combination of PSI CPU pressure and average CPU utilization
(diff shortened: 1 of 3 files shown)


pkg/apis/descheduler/v1/types_descheduler.go
```diff
@@ -129,6 +129,8 @@ const (
 	PrometheusCPUUsageProfile ActualUtilizationProfile = "PrometheusCPUUsage"
 	// PrometheusCPUPSIPressureProfile sets rate(node_pressure_cpu_waiting_seconds_total[1m]) query
 	PrometheusCPUPSIPressureProfile ActualUtilizationProfile = "PrometheusCPUPSIPressure"
+	// PrometheusCPUPSIPressureUtilizationProfile sets a query based on a combination of PSI CPU pressure and average CPU utilization
+	PrometheusCPUPSIPressureUtilizationProfile ActualUtilizationProfile = "PrometheusCPUPSIPressureUtilizationProfile"
```
ingvagabund, line 133: PrometheusCPUPSIPressureUtilizationProfile -> PrometheusCPUPSIPressureByUtilization

ingvagabund (changes requested):
One nit :)

PR #464, opened 2025-03-24
Introducing soft-tainter controller

tiraboschi:
Introducing a new controller to dynamically
set/remove a soft taint (effect: PreferNoSchedule)
to/from nodes that the descheduler identified
as overutilized according to the nodeutilization plugin.

Now the nodeutilization plugin in the descheduler
can consume utilization data from actual kubernetes metrics
and not only static reservations (requests).
On the other side, the default kube-scheduler
is only going to ensure that Pod's specific
resource requests can be satisfied.

The soft taint will simply gave an hint to the scheduler
to try avoiding nodes that looks overutilized at
descheduler eyes.
Since it's just a soft taint (do it just if possible)
there is no need to restrict it only to a certain
amount of nodes in the cluster or be strict on error handling.

This could potentially be directly implemented
in the descheduler but it has been judged as out of scope there.
So an external controller seems a more manageable solution.
Since we want to limit the rights to amend nodes to a
dedicated serviceaccount, the new controller will be executed
in a new pod deployed by the operator.
The operator will also deploy the ClusterRole
and the RoleBindings for this additional operand.

With a fine-grained ValidatingAdmissionPolicy (GA in k8s 1.30)
we can enforce that the softTainter service account
is only allowed to apply/remove soft-taints
(enforcing PreferNoSchedule effect)
and with a predefined key-prefix.
The operator will also deploy the ValidatingAdmissionPolicy.

Finally, a smarter node classification requires a more complex logic
that is implemented on Prometheous side with recording rules.
So the operator will now also deploy the PrometheusRule
checking that the containing namespace is correctly labelled
to exposed them.

Signed-off-by: Simone Tiraboschi <stirabos@redhat.com>


pkg/softtainter/nodeclassifier.go
```diff
@@ -0,0 +1,327 @@
...
+		func(nodeName string, usage, threshold api.ResourceThresholds) bool {
+			return isNodeAboveThreshold(usage, threshold)
+		},
+	)
+
+	// the nodeutilization package was designed to work with NodeInfo
+	// structs. these structs holds information about how utilized a node
+	// is. we need to go through the result of the classification and turn
+	// it into NodeInfo structs.
+	nodeInfos := make([][]nodeutilization.NodeInfo, 2)
```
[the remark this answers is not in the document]
tiraboschi: Done

bindata/assets/kube-descheduler/prometheusrule.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - name: recordingRules.rules
+      rules:
+        - record: descheduler:nodeutilization:cpu:avg1m
+          expr: avg by (instance) (1 - rate(node_cpu_seconds_total{mode='idle'}[1m]))
+
+        - record: descheduler:averageworkersutilization:cpu:avg1m
+          expr: avg(descheduler:nodeutilization:cpu:avg1m * on(instance) group_left(node) label_replace(kube_node_role{role="worker"}, 'instance', "$1", 'node', '(.+)'))
+
+        - record: descheduler:nodepressure:cpu:avg1m
+          expr: avg by (instance) ( rate(node_pressure_cpu_waiting_seconds_total[1m])) and (1 - avg by (instance) (rate(node_cpu_seconds_total{mode='idle'}[1m]))) > 0.7 or avg by (instance) ( rate(node_pressure_cpu_waiting_seconds_total[1m])) * 0
```
ricardomaraschini, line 18: Please see if my understanding of this is correct.

```suggestion
          # return the cpu pressure if the cpu usage is over 70% otherwise
          # return cpu pressure as zero.
          expr: |-
            avg by (instance) (
              rate(node_pressure_cpu_waiting_seconds_total[1m])
            ) and (
              1 - avg by (instance) (
                rate(node_cpu_seconds_total{mode='idle'}[1m])
              )
            ) > 0.7
            or
            avg by (instance) (
              rate(node_pressure_cpu_waiting_seconds_total[1m])
            ) * 0       
```

pkg/operator/target_config_reconciler.go
```diff
@@ -196,6 +202,16 @@ func (c TargetConfigReconciler) sync() error {
 		specAnnotations["serviceaccounts/openshift-descheduler-operand"] = resourceVersion
 	}
 
+	if stsa, _, err := c.manageSoftTainterServiceAccount(descheduler); err != nil {
+		return err
+	} else {
+		resourceVersion := "0"
+		if stsa != nil { // SyncConfigMap can return nil
```
ricardomaraschini, line 209: I think these comments regarding `SyncConfigMap` don't make sense in most of these.

pkg/operator/target_config_reconciler.go
```diff
@@ -1207,12 +1405,13 @@ func (c *TargetConfigReconciler) manageDeployment(descheduler *deschedulerv1.Kub
...
 
 	images := map[string]string{
-		"${IMAGE}": c.targetImagePullSpec,
+		"${IMAGE}": targetImagePullSpec,
 	}
 	for i := range required.Spec.Template.Spec.Containers {
 		for pat, img := range images {
 			if required.Spec.Template.Spec.Containers[i].Image == pat {
 				required.Spec.Template.Spec.Containers[i].Image = img
+				required.Spec.Template.Spec.Containers[i].ImagePullPolicy = v1.PullAlways // TODO: remove me
```
ricardomaraschini, line 1414: Maybe you forgot to remove this line (or the comment).

pkg/softtainter/node.go
```diff
@@ -0,0 +1,125 @@
...
+
+// IsNodeSoftTainted checks if the node is already tainted
+// with the provided key and value and PreferNoSchedule effect.
+// If a match is found, it returns true.
+func IsNodeSoftTainted(node *v1.Node, taintKey, taintValue string) bool {
+	if node == nil {
+		return false
+	}
+	for _, taint := range node.Spec.Taints {
+		if (taint.Effect == v1.TaintEffectPreferNoSchedule) && taint.Key == taintKey && taint.Value == taintValue {
```
ricardomaraschini, line 22:
```suggestion
		if taint.Effect == v1.TaintEffectPreferNoSchedule && taint.Key == taintKey && taint.Value == taintValue {
```

pkg/softtainter/node.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// patchNodeTaints patches node's taints.
+func patchNodeTaints(ctx context.Context, c client.Client, nodeName string, oldNode, newNode *v1.Node) error {
+	// Strip base diff node from RV to ensure that our Patch request will set RV to check for conflicts over .spec.taints.
+	// This is needed because .spec.taints does not specify patchMergeKey and patchStrategy and adding them is no longer an option for compatibility reasons.
+	// Using other Patch strategy works for adding new taints, however will not resolve problem with taint removal.
+	oldNodeNoRV := oldNode.DeepCopy()
+	oldNodeNoRV.ResourceVersion = ""
+	oldDataNoRV, err := json.Marshal(&oldNodeNoRV)
```
ricardomaraschini, line 101:
```suggestion
	oldDataNoRV, err := json.Marshal(oldNodeNoRV)
```

pkg/softtainter/nodeutilization.go
```diff
@@ -0,0 +1,224 @@
...
+		slices.Collect(maps.Keys(usage)),
+		[]api.ResourceThresholds{lowSpan, highSpan},
+	)
+	return usage, thresholds
+}
+
+// capNodeCapacitiesToThreshold caps the node capacities to the given
+// thresholds. if a threshold is not set for a resource, the full capacity is
+// returned.
+func capNodeCapacitiesToThreshold(
```
ricardomaraschini, line 67: This function seems to be unused.

pkg/softtainter/nodeutilization.go
```diff
@@ -0,0 +1,224 @@
...
+		capped[resourceName] = capNodeCapacityToThreshold(
+			node, thresholds, resourceName,
+		)
+	}
+	return capped
+}
+
+// capNodeCapacityToThreshold caps the node capacity to the given threshold. if
+// no threshold is set for the resource, the full capacity is returned.
+func capNodeCapacityToThreshold(
```
ricardomaraschini, line 83: This function seems to be unused.

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,507 @@
...
+func (st *softTainter) syncSoftTaints(ctx context.Context, nodes []*corev1.Node) error {
+	nodesMap := make(map[string]*corev1.Node, len(nodes))
+
+	nu := st.nodeUtilizationFactory(st.promClient, st.promQuery)
+	nodeUsages, err := nu.NodeUsageFromPrometheusMetrics(ctx)
+	if err != nil {
+		return err
+	}
+
+	// TODO: do we need this???
```
ricardomaraschini, line 209: You have a good question here. I guess that for the soft tainter with Prometheus the capacity will always be 100%.

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,507 @@
...
+				apprNodes = append(apprNodes, nodesMap[nodeName])
+			}
+		}
+	}
+	log.Info("Number of under utilized nodes", "totalNumber", len(lowNodes))
+	log.Info("Number of appropriately utilized nodes", "totalNumber", len(apprNodes))
+	log.Info("Number of over utilized nodes", "totalNumber", len(highNodes))
+
+	if st.args.mode == desv1.Automatic {
+		st.taintNodes(ctx, lowNodes, apprNodes, highNodes)
```
ricardomaraschini, line 321: `taintNodes()` may fail.

pkg/operator/target_config_reconciler.go
```diff
@@ -1008,6 +1204,11 @@ func (c *TargetConfigReconciler) manageConfigMap(descheduler *deschedulerv1.Kube
 		}
 	}
 
+	if descheduler.Spec.ProfileCustomizations != nil && descheduler.Spec.ProfileCustomizations.DevKubevirtSchedulable {
```
ingvagabund, line 1207: This needs to go under `relieveAndMigrateProfile` as it's relevant only to the new profile. Alternatively, you could introduce a separate `DevNodeSelector` tuning option for all plugins/profiles. Either as a string (to allow any node selector) or a predefined list of options (e.g. master nodes, worker nodes, kubevirtschedulable nodes`. Given we have not had any request from customers for this functionality it's better to have it scoped only for the profile for now. Also, would it make more sense to have this configuration by default without any tuning options? Are there are kubevirt use cases where it's better to also include nodes which does not have the label?

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,507 @@
...
+			return isNodeAboveThreshold(usage, threshold)
+		},
+	)
+	var lowNodes []*corev1.Node
+	var apprNodes []*corev1.Node
+	var highNodes []*corev1.Node
+	categories := []string{"underutilized", "overutilized"}
+
+	classifiedNodes := map[string]bool{}
+	for i := range nodeGroups {
```
ricardomaraschini, line 273: `lowNodes`, `highNodes` and `apprNodes` slices may all be populated inside the classifiers. To classify the `apprNodes` you only need a third classifier that returns always `true`. I think this will simplify a lot here.

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,507 @@
...
+		for k, v := range map[string]string{
+			appropriatelySoftTaintKey: appropriatelyutilizedSoftTaintValue,
+		} {
+			err := st.addTaint(ctx, node, k, v)
+			if err != nil {
+				return err
+			}
+		}
+		for k, v := range map[string]string{
+			overutilizedSoftTaintKey: overutilizedSoftTaintValue,
```
ricardomaraschini, line 412: Shouldn't we be dropping the under utilized soft taint key as well ?

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,507 @@
...
+			}
+		}
+	}
+	if lnargs == nil {
+		err := fmt.Errorf("unable to read LowNodeUtilizationArgs")
+		logger.Error(err, "reconciliation failed")
+		return reconcile.Result{}, err
+	}
+
+	enableSoftTainter := false
```
ingvagabund, line 125: `enableSoftTainter := des.Spec.ProfileCustomizations != nil && des.Spec.ProfileCustomizations.DevEnableSoftTainter`

ingvagabund: Assuming `DevEnableSoftTainter` means soft-taint nodes it's better to avoid rendering the controller deployment completely when `DevEnableSoftTainter` is false. Or, is there any benefit of running the controller without tainting the nodes besides removing existing taints?

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,507 @@
...
+
+	if des.Spec.DeschedulingIntervalSeconds == nil {
+		return reconcile.Result{}, fmt.Errorf("descheduler should have an interval set")
+	}
+	st.resyncPeriod = time.Duration(*des.Spec.DeschedulingIntervalSeconds) * time.Second
+
+	var lnargs *nodeutilization.LowNodeUtilizationArgs
+
+	for _, p := range policy.Profiles {
+		if p.Name == string(desv1.LongLifecycle) || p.Name == string(desv1.RelieveAndMigrate) {
```
ingvagabund, line 111: We can limit running the soft tainter only for `RelieveAndMigrate` profile. Everyone who wants to run the new controller will be asked to set the new profile. Including anyone using the `LongLifecycle` profile. So we limit the support and encourage users to adopt the new profile.

pkg/operator/target_config_reconciler_test.go
```diff
@@ -237,6 +238,37 @@ func TestManageConfigMap(t *testing.T) {
 				Data:     map[string]string{"policy.yaml": string(bindata.MustAsset("assets/relieveAndMigrateMediumConfig.yaml"))},
 			},
 		},
+		{
+			name: "RelieveAndMigrateDeviationMedium",
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/operator/target_config_reconciler_test.go
```diff
@@ -865,6 +905,160 @@ func TestManageDeployment(t *testing.T) {
 	}
 }
 
+func TestManagesoftTainterDeployment(t *testing.T) {
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/softtainter/node.go
tiraboschi: done

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,507 @@
...
+	err = st.client.List(ctx, &nl, &client.ListOptions{LabelSelector: nodeSelector})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	nodeList := make([]*corev1.Node, len(nl.Items))
+	for i, node := range nl.Items {
+		nodeList[i] = &node
+	}
+
+	if st.args.applySoftTaints {
```
ingvagabund, line 191: This indicates the soft taints are removed only when `applySoftTaints` is true. When the operator is uninstalled there's not much we can do as owner reference can take care of removing resources/manifests owned by the operator. Yet, the node taints need to be removed explicitly. Something to mention in the readme and documentation. 

When the soft tainter is disabled all the existing soft taints need to be removed. Which could be handled by the operator deploying the soft taint controller in "cleaning" mode and waiting until all soft taints are removed. Then scaling the deployment to 0. Or, removing the deployment (and corresponding resources like SA, VAP, ...).

pkg/softtainter/nodeutilization.go
tiraboschi: done

pkg/softtainter/nodeutilization.go
tiraboschi: done

pkg/softtainter/softtainter.go
tiraboschi: done

pkg/softtainter/softtainter.go
tiraboschi: No, I think we can completely skip it to save resources

pkg/softtainter/softtainter.go
tiraboschi: Likely yes, but I'm not 100% convinced because maybe in the future we can image other queries that could require something like this

pkg/apis/descheduler/v1/types_descheduler.go
```diff
@@ -71,10 +71,20 @@ type ProfileCustomizations struct {
...
 
-	// LowNodeUtilizationThresholds enumerates predefined experimental thresholds
+	// DevLowNodeUtilizationThresholds enumerates predefined experimental thresholds
 	// +kubebuilder:validation:Enum=Low;Medium;High;""
 	DevLowNodeUtilizationThresholds *LowNodeUtilizationThresholdsType `json:"devLowNodeUtilizationThresholds"`
 
+	// DevEnableSoftTainter enables SoftTainter alpha feature.
+	// The EvictionsInBackground alpha feature is a subject to change.
+	// Currently provided as an experimental feature.
+	DevEnableSoftTainter bool `json:"devEnableSoftTainter"`
```
ingvagabund, line 81: Would you please extend the README.md with the new tuning options as well? `DevEnableSoftTainter` is a good example of disabling the controller in case users decide to provide their own implementation. Yet, still limited in the scope of the new profile only.

pkg/softtainter/softtainter.go
tiraboschi: I was thinking about that.
But I also see an additional case that I'm not sure we can solve in that way.
Now the SoftTainter is removing all the softtaints from under utilized nodes, adding one to appropriately utilized and two to overutilized.
When we have dynamic thresholds and the system is converging, all the nodes are ideally expected to be appropriately utilized. But in that case we do not want to set any soft taint.
That logic is actually here and i'm not sure I can replicate it with an additional classifier.

pkg/softtainter/softtainter.go
tiraboschi: done

pkg/softtainter/node.go
```diff
@@ -0,0 +1,125 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// IsNodeSoftTainted checks if the node is already tainted
+// with the provided key and value and PreferNoSchedule effect.
+// If a match is found, it returns true.
+func IsNodeSoftTainted(node *v1.Node, taintKey, taintValue string) bool {
```
ingvagabund, line 17: [k8s.io/kubernetes/pkg/util/taints](https://github.com/kubernetes/kubernetes/blob/master/pkg/util/taints/taints.go) has useful functions you could utilize here. E.g. `AddOrUpdateTaint`, `AddOrUpdateTaint` or `TaintExists`.

pkg/operator/target_config_reconciler.go
```diff
@@ -64,22 +65,25 @@ const DefaultImage = "quay.io/openshift/origin-descheduler:latest"
...
+	configSchedulerLister    configlistersv1.SchedulerLister
+	routeRouteLister         routelistersv1.RouteLister
+	cache                    resourceapply.ResourceCache
 }
 
 func NewTargetConfigReconciler(
 	ctx context.Context,
-	targetImagePullSpec string,
+	deschedulerImagePullSpec string,
+	softtainterImagePullSpec string,
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/operator/target_config_reconciler.go
tiraboschi: done

pkg/softtainter/softtainter.go
ricardomaraschini: Do you mean doing something like this ? Just checking if I understood.

```go
if len(low) == 0 {
    low = appropriate
    appropriate = []
}
```

pkg/operator/target_config_reconciler.go
```diff
@@ -366,6 +539,23 @@ func (c *TargetConfigReconciler) manageServiceAccount(descheduler *deschedulerv1
 	return resourceapply.ApplyServiceAccount(c.ctx, c.kubeClient.CoreV1(), c.eventRecorder, required)
 }
 
+func (c *TargetConfigReconciler) manageSoftTainterServiceAccount(descheduler *deschedulerv1.KubeDescheduler) (*v1.ServiceAccount, bool, error) {
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/softtainter/softtainter.go
tiraboschi: done

pkg/softtainter/softtainter.go
tiraboschi: done

pkg/softtainter/node.go
tiraboschi: I took `TaintExists`, `AddOrUpdateTaint` is less useful here since it will amend the node object node but here i'm using a patch operator to prevent conflicts.

pkg/softtainter/softtainter.go
tiraboschi: I tried adding the third, catch all, classifier at the end of the list, but i'm not sure it's ever called.

pkg/apis/descheduler/v1/types_descheduler.go
tiraboschi: Done

ingvagabund:
You will also need to extend https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/cluster-kube-descheduler-operator.clusterserviceversion.yaml#L139 with any new rbac that's getting added through bindata/assets/kube-descheduler. As an operator needs to have the same permissions it gives to its operands/components it reconciles.

tiraboschi:
> You will also need to extend https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/cluster-kube-descheduler-operator.clusterserviceversion.yaml#L139 with any new rbac that's getting added through `bindata/assets/kube-descheduler`. As an operator needs to have/possess the same permissions it gives to its operands/components it reconciles.

I think that a better solution is simply having the OLM managing the softtainter serviceaccount and its clusterrole for us

tiraboschi:
> You will also need to extend https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/cluster-kube-descheduler-operator.clusterserviceversion.yaml#L139 with any new rbac that's getting added through `bindata/assets/kube-descheduler`. As an operator needs to have/possess the same permissions it gives to its operands/components it reconciles.

done

README.md
```diff
@@ -200,6 +201,7 @@ the `profileCustomizations` field:
 | `devHighNodeUtilizationThresholds` | `string` | Sets thresholds for the [HighNodeUtilization](https://github.com/kubernetes-sigs/descheduler#highnodeutilization) strategy of the `CompactAndScale` profile in the following ratios: `Minimal` for 10%, `Modest` for 20%, `Moderate` for 30%. Currently provided as an experimental feature.|
 |`devActualUtilizationProfile`|`string`| Sets a profile that gets translated into a predefined prometheus query |
 | `devDeviationThresholds` | `string` | Have the thresholds be based on the average utilization. Thresholds signify the distance from the average node utilization in the following setting: `Low`: 10%:10%, `Medium`: 20%:20%, `High`: 30%:30% |
```
ingvagabund, line 203: Can you also please extend the list here with `AsymmetricLow`, `AsymmetricMedium` and `AsymmetricHigh`?

pkg/operator/starter.go
```diff
@@ -99,6 +99,7 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	targetConfigReconciler := NewTargetConfigReconciler(
 		ctx,
 		os.Getenv("RELATED_IMAGE_OPERAND_IMAGE"),
+		os.Getenv("RELATED_IMAGE_SOFTTAINTER_IMAGE"),
```
ingvagabund, line 102: This requires to update bundle.Dockerfile with:
```
ARG SOFTTAINER_IMAGE=registry.stage.redhat.io/kube-descheduler-operator/kube-descheduler-rhel9-operator@sha256:426baca9e36c7833d335262a7b8f73c318cd749ca7767726fc86109948a44d96
ARG REPLACED_SOFTTAINER_IMG=quay.io/openshift/origin-cluster-kube-descheduler-operator:latest
```
and set `RELATED_IMAGE_SOFTTAINTER_IMAGE` under the csv file (in the deployment definition the same way as in `deploy/05_deployment.yaml`).

deploy/05_deployment.yaml
```diff
@@ -42,6 +42,8 @@ spec:
               value: "descheduler-operator"
             - name: RELATED_IMAGE_OPERAND_IMAGE
               value: quay.io/openshift/origin-descheduler:latest
+            - name: RELATED_IMAGE_SOFTTAINTER_IMAGE
```
ingvagabund, line 45: FYI: manifests under `deploy` directory are used only for testing purposes.

pkg/operator/target_config_reconciler.go
```diff
@@ -1287,3 +1540,43 @@ func (c *TargetConfigReconciler) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
 	}
 }
+
+func (c *TargetConfigReconciler) checkNamepsaceMonitoringLabel() error {
+	operatorNamespace, err := c.kubeClient.CoreV1().Namespaces().Get(c.ctx, operatorclient.OperatorNamespace, metav1.GetOptions{})
```
ingvagabund, line 1545: This needs to be changed to read the namespace from the informer, not directly accessing the apiserver to minimize the traffic.

pkg/operator/target_config_reconciler.go
```diff
@@ -1287,3 +1540,43 @@ func (c *TargetConfigReconciler) eventHandler() cache.ResourceEventHandler {
...
+		return fmt.Errorf("namespace %v is not labeled with %v=%v", operatorclient.OperatorNamespace, operatorclient.OpenshiftClusterMonitoringLabelKey, operatorclient.OpenshiftClusterMonitoringLabelValue)
+	}
+	return nil
+}
+
+func (c *TargetConfigReconciler) isSoftTainterNeeded(descheduler *deschedulerv1.KubeDescheduler) (bool, error) {
+	if descheduler.Spec.ProfileCustomizations != nil && descheduler.Spec.ProfileCustomizations.DevEnableSoftTainter {
+		return true, nil
+	}
+	nodes, err := c.kubeClient.CoreV1().Nodes().List(c.ctx, metav1.ListOptions{})
```
ingvagabund, line 1560: This needs to be changed to read the nodes from the informer, not directly accessing the apiserver to minimize the traffic.

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,532 @@
...
+	if policy.NodeSelector != nil {
+		sel, err := labels.Parse(*policy.NodeSelector)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		nodeSelector = sel
+	}
+
+	nl := corev1.NodeList{}
+	err = st.client.List(ctx, &nl, &client.ListOptions{LabelSelector: nodeSelector})
```
ingvagabund, line 183: Is `List` getting listed from an informer cache or directly from a cluster? The former is required to reduce the traffic between the controller and the apiserver.

README.md
tiraboschi: done

pkg/operator/target_config_reconciler.go
```diff
@@ -1046,6 +1282,8 @@ func (c *TargetConfigReconciler) manageConfigMap(descheduler *deschedulerv1.Kube
 		case deschedulerv1.CompactAndScale:
 			profile, err = compactAndScaleProfile(descheduler.Spec.ProfileCustomizations, includedNamespaces, excludedNamespaces, ignorePVCPods, evictLocalStoragePods)
 		case deschedulerv1.RelieveAndMigrate:
+			kubeVirtShedulable := "kubevirt.io/schedulable=true"
+			policy.NodeSelector = &kubeVirtShedulable
```
ingvagabund, line 1286: Can you also mention this in the README? So users are aware the node selector is always set when the profile is used.

pkg/operator/target_config_reconciler.go
```diff
@@ -1207,12 +1405,13 @@ func (c *TargetConfigReconciler) manageDeployment(descheduler *deschedulerv1.Kub
 	}
 
 	images := map[string]string{
-		"${IMAGE}": c.targetImagePullSpec,
+		"${IMAGE}": targetImagePullSpec,
```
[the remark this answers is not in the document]
ingvagabund: This is still valid. Yet, not strictly required. I will leave it up to you.

pkg/softtainter/softtainter.go
tiraboschi: Controller-runtime is client is a caching client (internally based on informers for registered kinds): all the operations are actually performed against a cache.
See `func getCacheOption` in `cmd/soft-tainter/main.go`.

pkg/operator/target_config_reconciler.go
ingvagabund: This one is still valid

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,532 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	prometheusAuthTokenSecretKey        = "prometheusAuthToken"
+	OverUtilizedSoftTaintKey            = "nodeutilization.descheduler.kubernetes.io/overutilized"
```
ingvagabund, line 39: nodeutilization.descheduler.kubernetes.io -> nodeutilization.descheduler.sigs.k8s.io

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,532 @@
...
+	OverUtilizedSoftTaintValue          = "true"
+	AppropriatelyUtilizedSoftTaintKey   = "nodeutilization.descheduler.kubernetes.io/appropriate"
```
ingvagabund, line 41: ditto

pkg/softtainter/softtainter.go
ingvagabund: Compared to https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions we might use `nodeutilization.descheduler.sigs.kubernetes.io`?

ingvagabund: Or, given this is a TP/beta feature what about `nodeutilization.descheduler.beta.kubernetes.io`?

ingvagabund: Also, given this is an openshift specific taint we might go with `nodeutilization.descheduler.beta.openshift.io`?

tiraboschi: Version dependent taint names will require a leftover cleanup step on upgrades.
If we don't see any good reason to make them version specific, I'd suggest to go with `nodeutilization.descheduler.openshift.io`.

pkg/operator/target_config_reconciler.go
tiraboschi: Done

pkg/operator/starter.go
tiraboschi: done

pkg/softtainter/softtainter.go
ingvagabund: I see. Let's go with `nodeutilization.descheduler.openshift.io` then.

pkg/operator/target_config_reconciler.go
tiraboschi: Done

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,532 @@
...
+	resyncPeriod                      time.Duration
+	policyConfigFile                  string
+	nodeUtilizationFactory            func(promapi.Client, string) NodeUtilization
+}
+
+type NodeUtilization interface {
+	NodeUsageFromPrometheusMetrics(ctx context.Context) (map[string]map[corev1.ResourceName]*resource.Quantity, error)
+}
+
+type ActualNodeUtilization struct {
```
ingvagabund, line 75: Is this type used outside of the package? If not let's keep it private to avoid any extra support&maintenance.

pkg/softtainter/softtainter.go
tiraboschi: Done

pkg/operator/target_config_reconciler.go
tiraboschi: done

pkg/operator/target_config_reconciler.go
tiraboschi: done

pkg/softtainter/softtainter.go
```diff
@@ -0,0 +1,532 @@
...
+	promClient promapi.Client
+	promQuery  string
+}
+
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/softtainter/softtainter.go
tiraboschi: done

pkg/operator/target_config_reconciler.go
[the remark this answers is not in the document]
tiraboschi: done

pkg/softtainter/softtainter.go
ricardomaraschini: I guess we are OK with this then.

README.md
```diff
@@ -199,7 +203,8 @@ the `profileCustomizations` field:
...
 |`devActualUtilizationProfile`|`string`| Sets a profile that gets translated into a predefined prometheus query |
-| `devDeviationThresholds` | `string` | Have the thresholds be based on the average utilization. Thresholds signify the distance from the average node utilization in the following setting: `Low`: 10%:10%, `Medium`: 20%:20%, `High`: 30%:30% |
+| `devDeviationThresholds` | `string` | Have the thresholds be based on the average utilization. Thresholds signify the distance from the average node utilization. An AsymmetricDeviationThreshold will force all nodes below the average to be considered as underutilized to help rebalancing overutilized outliers. Supported settings: `Low`: 10%:10%, `Medium`: 20%:20%, `High`: 30%:30%, `AsymmetricLow`: 0%:10%, `AsymmetricMedium`: 0%:20%, `AsymmetricHigh`: 0%:30% |
+| `devEnableSoftTainter` | `bool` | Have the operator deploying the soft-tainter component to dynamic set/remove soft taints a a hint for the scheduler based on the same criteria used for the descheduling |
```
ricardomaraschini, line 207:
```suggestion
| `devEnableSoftTainter` | `bool` | Have the operator deploying the soft-tainter component to dynamic set/remove soft taints as a hint for the scheduler based on the same criteria used for the descheduling |
```

bindata/assets/kube-descheduler/prometheusrule.yaml
```diff
@@ -0,0 +1,39 @@
...
+                rate(node_cpu_seconds_total{mode='idle'}[1m])
+              )
+            ) > 0.7
+            or
+            avg by (instance) (
+              rate(node_pressure_cpu_waiting_seconds_total[1m])
+            ) * 0    
+
+        - record: descheduler:combined_utilization_and_pressure:avg1m
+          expr: |
```
ricardomaraschini, line 36: For consistency.

```suggestion
          expr: |-
```

pkg/softtainter/softtainter.go
tiraboschi: I'll amend the comment to keep track of it

bindata/assets/kube-descheduler/softtainterclusterrole.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - "update"
+  - "delete"
+- apiGroups:
+    - "operator.openshift.io"
+  resources:
+    - "kubedeschedulers"
+  verbs:
+    - "get"
+    - "watch"
+    - "list"
```
ricardomaraschini, line 52: Uber nit: can we fix the indentation of these arrays ?

bindata/assets/kube-descheduler/softtaintervalidatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,52 @@
...
+    objectSelector: {}
+    resourceRules:
+      - apiGroups:   [""]
+        apiVersions: ["*"]
+        operations:  ["UPDATE"]
+        resources:   ["nodes"]
+        scope: "*"
+  matchConditions:
+    - name: 'openshift-descheduler-softtainter'
+      expression: "request.userInfo.username=='system:serviceaccount:openshift-kube-descheduler-operator:openshift-descheduler-softtainter' || request.userInfo.username=='system:serviceaccount:openshift-kube-descheduler-operator:openshift-descheduler'"
```
ricardomaraschini, line 19: Any reason why we also included the `openshift-descheduler` service account here ?

pkg/operator/resourceapply_missing.go
```diff
@@ -0,0 +1,39 @@
...
+	err := client.ValidatingAdmissionPolicies().Delete(ctx, required.Name, metav1.DeleteOptions{})
+	if err != nil && apierrors.IsNotFound(err) {
+		return nil, false, nil
+	}
+	if err != nil {
+		return nil, false, err
+	}
+	resourcehelper.ReportDeleteEvent(recorder, required, err)
+	return nil, true, nil
+}
```
ricardomaraschini, line 27: This (and the next function) always have `nil` as the first returned value.  Is that to keep it compatible with what we currently do in `library-go` ?

bindata/assets/kube-descheduler/softtaintervalidatingadmissionpolicy.yaml
tiraboschi: With k8s RBAC rules we can only grant EDIT/PATCH rights at object level (nodes). We are using this `ValidatingAdmissionPolicy` to be more fine grained limiting it to soft-tatins.
This is needed for the service account used by the soft-tainter and also for the operator one since now it also needs node edit rights in order to be allowed to propagate them to the soft-tainter account that is dynamically created by the operator (only if the soft-tainter feature is enabled).
Descheduler service-account is not getting node edit rights so no need to limit it with the VAP.

pkg/operator/resourceapply_missing.go
tiraboschi: Exactly, check for instance: https://github.com/openshift/library-go/blob/ce2ba53fb2a4b7318769480e897a0e506b7cb0c0/pkg/operator/resource/resourceapply/rbac.go#L206-L252 or https://github.com/openshift/library-go/blob/ce2ba53fb2a4b7318769480e897a0e506b7cb0c0/pkg/operator/resource/resourceapply/core.go#L628-L698 
Honestly I'm also not a big fan of that but I understand that keeping the same signatures between create and delete functions is a nice to have.

bindata/assets/kube-descheduler/softtaintervalidatingadmissionpolicy.yaml
ricardomaraschini: TIL.

bindata/assets/kube-descheduler/prometheusrule.yaml
tiraboschi: done

bindata/assets/kube-descheduler/softtainterclusterrole.yaml
tiraboschi: done

README.md
tiraboschi: done

pkg/operator/resourceapply_missing.go
ricardomaraschini: OK, all good.

pkg/operator/target_config_reconciler.go
```diff
@@ -261,9 +361,15 @@ func (c TargetConfigReconciler) sync() error {
 			Status: operatorv1.ConditionFalse,
 		}),
 		func(status *operatorv1.OperatorStatus) error {
-			resourcemerge.SetDeploymentGeneration(&status.Generations, deployment)
+			resourcemerge.SetDeploymentGeneration(&status.Generations, deschedulerDeployment)
 			return nil
-		})
+		},
+		// TODO: do it only if enabled
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/operator/target_config_reconciler.go
```diff
@@ -1008,6 +1249,12 @@ func (c *TargetConfigReconciler) manageConfigMap(descheduler *deschedulerv1.Kube
 		}
 	}
 
+	if descheduler.Spec.ProfileCustomizations != nil && descheduler.Spec.ProfileCustomizations.DevEnableSoftTainter {
+		if !slices.Contains(descheduler.Spec.Profiles, deschedulerv1.RelieveAndMigrate) {
+			return nil, true, fmt.Errorf("the softtainter can only be used with the RelieveAndMigrate profile")
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/operator/target_config_reconciler.go
```diff
@@ -1046,6 +1293,8 @@ func (c *TargetConfigReconciler) manageConfigMap(descheduler *deschedulerv1.Kube
 		case deschedulerv1.CompactAndScale:
 			profile, err = compactAndScaleProfile(descheduler.Spec.ProfileCustomizations, includedNamespaces, excludedNamespaces, ignorePVCPods, evictLocalStoragePods)
 		case deschedulerv1.RelieveAndMigrate:
+			kubeVirtShedulable := kubeVirtShedulableLabelSelector
+			policy.NodeSelector = &kubeVirtShedulable
```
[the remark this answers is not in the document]
tiraboschi: done

README.md
```diff
@@ -199,7 +203,8 @@ the `profileCustomizations` field:
...
 |`devActualUtilizationProfile`|`string`| Sets a profile that gets translated into a predefined prometheus query |
-| `devDeviationThresholds` | `string` | Have the thresholds be based on the average utilization. Thresholds signify the distance from the average node utilization in the following setting: `Low`: 10%:10%, `Medium`: 20%:20%, `High`: 30%:30% |
+| `devDeviationThresholds` | `string` | Have the thresholds be based on the average utilization. Thresholds signify the distance from the average node utilization. An AsymmetricDeviationThreshold will force all nodes below the average to be considered as underutilized to help rebalancing overutilized outliers. Supported settings: `Low`: 10%:10%, `Medium`: 20%:20%, `High`: 30%:30%, `AsymmetricLow`: 0%:10%, `AsymmetricMedium`: 0%:20%, `AsymmetricHigh`: 0%:30% |
+| `devEnableSoftTainter` | `bool` | Have the operator deploying the soft-tainter component to dynamic set/remove soft taints as a hint for the scheduler based on the same criteria used for the descheduling |
```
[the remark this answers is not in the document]
tiraboschi: done

bundle.Dockerfile
```diff
@@ -9,9 +9,12 @@ ARG OPERATOR_IMAGE=registry.stage.redhat.io/kube-descheduler-operator/kube-desch
 ARG OPERAND_IMAGE=registry.stage.redhat.io/kube-descheduler-operator/descheduler-rhel9@sha256:e7f1e09ea814190f9c08c7e6a233d1008c937116b278ef92e9de18bcbf10807e
 ARG REPLACED_OPERATOR_IMG=registry-proxy.engineering.redhat.com/rh-osbs/kube-descheduler-operator-rhel-9:latest
 ARG REPLACED_OPERAND_IMG=registry-proxy.engineering.redhat.com/rh-osbs/descheduler-rhel-9:latest
+ARG SOFTTAINER_IMAGE=registry.stage.redhat.io/kube-descheduler-operator/kube-descheduler-rhel9-operator@sha256:426baca9e36c7833d335262a7b8f73c318cd749ca7767726fc86109948a44d96
+ARG REPLACED_SOFTTAINER_IMG=quay.io/openshift/origin-cluster-kube-descheduler-operator:latest
```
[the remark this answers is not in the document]
tiraboschi: done

pkg/operator/target_config_reconciler.go
```diff
@@ -1241,6 +1487,25 @@ func (c *TargetConfigReconciler) manageDeployment(descheduler *deschedulerv1.Kub
...
+	return c.manageDeployment(required, descheduler, targetImageKey, c.deschedulerImagePullSpec, specAnnotations)
+}
+
+func (c *TargetConfigReconciler) manageSoftTainterDeployment(descheduler *deschedulerv1.KubeDescheduler, specAnnotations map[string]string, stEnabled bool) (*appsv1.Deployment, bool, error) {
+	const targetImageKey = "${SOFTTAINTER_IMAGE}"
+	required := resourceread.ReadDeploymentV1OrDie(bindata.MustAsset("assets/kube-descheduler/softtainterdeployment.yaml"))
+	required.Name = operatorclient.SoftTainterOperandName
+	required.Namespace = descheduler.Namespace
+	if stEnabled {
+		return c.manageDeployment(required, descheduler, targetImageKey, c.softtainterImagePullSpec, specAnnotations)
```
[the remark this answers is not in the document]
tiraboschi: done

PR #467, opened 2025-03-31
Extend the list of explictly protected namespaces with openshift namespace

ingvagabund:
This namespaces ("openshift") is created by default as well. It needs to be protected.


ingvagabund:
The e2e has not additional test for the new namespaces.

/override ci/prow/e2e-aws-operator

PR #469, opened 2025-04-01
bump descheduler: consume up to date descheduler

tiraboschi:
Consume up to date upstream descheduler code.
Fix an API change in MetricsUtilization and
Prometheus APIs.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_1.22 as builder
```
ingvagabund, line 1: Let's keep 1.22 for now. All the code changes will get backported to 4.18 which still needs 1.22.

ricardomaraschini (approved):

PR #506, opened 2025-04-17
Have the SoftTainter always deployed with KubeVirtRelieveAndMigrate

tiraboschi:
The SoftTainter is a critical mandatory
component for the correct behaviour of the
KubeVirtRelieveAndMigrate profile.
Have it always deployed when KubeVirtRelieveAndMigrate
is in use.
Deprecate DevEnableSoftTainter.

Fixes: https://issues.redhat.com/browse/WRKLDS-1673

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,11 +174,16 @@ That nodeSelector is a top level configuration option affecting all the active d
 this profile is not expected to be combined with other profiles
 unless all profiles are expected to operate over the same set of nodes.
 
+This profile deploys the soft-tainter component to dynamically set/remove soft taints according to the
+same criteria used for load aware descheduling. In case of load-aware descheduling we can potentially have a
+relevant asymmetry between the descheduling and successive scheduling decisions.
+The soft taints set by the descheduler soft-tainter act as a hint for the scheduler to mitigate
+this asymmetry and foster a quicker convergence.
+
 The profile exposes the following customization:
 - `devLowNodeUtilizationThresholds`: Sets experimental thresholds for the LowNodeUtilization strategy.
 - `devActualUtilizationProfile`: Enable load-aware descheduling.
 - `devDeviationThresholds`: Have the thresholds be based on the average utilization.
-- `devEnableSoftTainter`: Have the operator deploying the soft-tainter component to dynamically set/remove soft taints according to the same criteria used for load aware descheduling. Scheduling and descheduling decisions are in general fully independent. In case of load-aware descheduling we can potentially have a relevant asymmetry between the descheduling and successive scheduling decisions. The soft taints set by the descheduler soft-tainter act as a hint for the scheduler to mitigate this asymmetry and foster a quicker convergence.
 
 ### EvictPodsWithPVC
 By default, the operator prevents pods with PVCs from being evicted. Enabling this
@@ -207,7 +212,6 @@ the `profileCustomizations` field:
 | `devHighNodeUtilizationThresholds` | `string` | Sets thresholds for the [HighNodeUtilization](https://github.com/kubernetes-sigs/descheduler#highnodeutilization) strategy of the `CompactAndScale` profile in the following ratios: `Minimal` for 10%, `Modest` for 20%, `Moderate` for 30%. Currently provided as an experimental feature.|
 |`devActualUtilizationProfile`|`string`| Sets a profile that gets translated into a predefined prometheus query |
 | `devDeviationThresholds` | `string` | Have the thresholds be based on the average utilization. Thresholds signify the distance from the average node utilization. An AsymmetricDeviationThreshold will force all nodes below the average to be considered as underutilized to help rebalancing overutilized outliers. Supported settings: `Low`: 10%:10%, `Medium`: 20%:20%, `High`: 30%:30%, `AsymmetricLow`: 0%:10%, `AsymmetricMedium`: 0%:20%, `AsymmetricHigh`: 0%:30% |
-| `devEnableSoftTainter` | `bool` | Have the operator deploying the soft-tainter component to dynamic set/remove soft taints as a hint for the scheduler based on the same criteria used for the descheduling |
 
 ## Prometheus query profiles
 The operator provides the following profiles:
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -174,11 +174,18 @@ That nodeSelector is a top level configuration option affecting all the active d
 this profile is not expected to be combined with other profiles
 unless all profiles are expected to operate over the same set of nodes.
 
+This profiles, by default, deploys the soft-tainter component to dynamically set/remove soft taints according to the
```
ricardomaraschini, line 177:
```suggestion
This profile, by default, deploys the soft-tainter component to dynamically set/remove soft taints according to the
```

README.md
```diff
@@ -174,11 +174,18 @@ That nodeSelector is a top level configuration option affecting all the active d
...
+this asymmetry and foster a quicker convergence.
+The SoftTainter component can be eventually disabled with the KubevirtRelieveAndMigrateSoftTainter Profile Customizations.
+
 The profile exposes the following customization:
 - `devLowNodeUtilizationThresholds`: Sets experimental thresholds for the LowNodeUtilization strategy.
 - `devActualUtilizationProfile`: Enable load-aware descheduling.
 - `devDeviationThresholds`: Have the thresholds be based on the average utilization.
-- `devEnableSoftTainter`: Have the operator deploying the soft-tainter component to dynamically set/remove soft taints according to the same criteria used for load aware descheduling. Scheduling and descheduling decisions are in general fully independent. In case of load-aware descheduling we can potentially have a relevant asymmetry between the descheduling and successive scheduling decisions. The soft taints set by the descheduler soft-tainter act as a hint for the scheduler to mitigate this asymmetry and foster a quicker convergence.
+- `KubevirtRelieveAndMigrateSoftTainter`: Control if the profile will trigger the deployment of the soft-tainter component. Enabled by default.
```
ricardomaraschini, line 188: All the other customizations are prefixed with `dev`, to drop the prefix for this one is by design ?

README.md
```diff
@@ -207,7 +214,7 @@ the `profileCustomizations` field:
...
 |`devActualUtilizationProfile`|`string`| Sets a profile that gets translated into a predefined prometheus query |
 | `devDeviationThresholds` | `string` | Have the thresholds be based on the average utilization. Thresholds signify the distance from the average node utilization. An AsymmetricDeviationThreshold will force all nodes below the average to be considered as underutilized to help rebalancing overutilized outliers. Supported settings: `Low`: 10%:10%, `Medium`: 20%:20%, `High`: 30%:30%, `AsymmetricLow`: 0%:10%, `AsymmetricMedium`: 0%:20%, `AsymmetricHigh`: 0%:30% |
-| `devEnableSoftTainter` | `bool` | Have the operator deploying the soft-tainter component to dynamic set/remove soft taints as a hint for the scheduler based on the same criteria used for the descheduling |
+| `KubevirtRelieveAndMigrateSoftTainter` | `string` | Control if the KubeVirtRelieveAndMigrate profile will trigger the deployment of the soft-tainter component. Enabled by default. Not relevat for other profiles. |
```
ricardomaraschini, line 217: Wouldn't this be easier to reason about  with a bool named `DisableKubevirtRelieveAnMigrateSoftTainter`. If not specified then it is `false` by default. I have the impression that this would simplify things here as no extra type would be needed.

pkg/operator/target_config_reconciler.go
```diff
@@ -1254,7 +1254,7 @@ func (c *TargetConfigReconciler) manageConfigMap(descheduler *deschedulerv1.Kube
 		}
 	}
 
-	if descheduler.Spec.ProfileCustomizations != nil && descheduler.Spec.ProfileCustomizations.DevEnableSoftTainter {
+	if descheduler.Spec.ProfileCustomizations != nil && descheduler.Spec.ProfileCustomizations.KubevirtRelieveAndMigrateSoftTainter == deschedulerv1.SoftTainterEnabled {
```
ricardomaraschini, line 1257: Are we going to simply ignore the now deprecated `DevEnableSoftTainter` parameter ?

README.md
tiraboschi: @ingvagabund what do you suggest?

pkg/operator/target_config_reconciler.go
ricardomaraschini: You are absolutely correct here. I haven't noticed this caveat. We will have a parameter that does nothing, that is unfortunate (albeit acceptable IMHO as no one is using it yet).

| DevEnableSoftTainter | KubevirtRelieveAndMigrateSoftTainter  | Action |
| ----- | --- | --- |
| false | "Enabled" | Enables the Soft Tainter |
| true | "Enabled" | Enables the Soft Tainter |
| false  | "Disabled" | Disables the Soft Tainter |
| true  | "Disabled" | Disables the Soft Tainter |
| false | "" | **Not clear what is expected** |
| true  | "" | Enables the Soft Tainter |

README.md
ricardomaraschini: That is OK. I think you have a typo here, it should be `kubevirtRelieveAndMigrateSoftTainter` with a lower case.

pkg/operator/target_config_reconciler.go
tiraboschi: `DevEnableSoftTainter` is basically ignored, if `KubevirtRelieveAndMigrateSoftTainter` is not specified, the softtainter will be enabled (default option with the RelieveAndMigrate profile).

README.md
tiraboschi: done

pkg/operator/target_config_reconciler.go
ingvagabund: `DevEnableSoftTainter` is an override for case when the soft-tainter is disabled by default. Given it's gonna by enabled by default the field is no-op from now on. So it can be safely ignored.

ingvagabund:
Thinking about this again given the soft-tainter will be enabled by default we don't need to allow customers to disable it unless there's already a customer use cases asking for it. Let's introduce only if a customer asks for it.

PR #507, opened 2025-04-17
Add an alert and an condition for RelieveAndMigrate without PSI

tiraboschi:
RelieveAndMigrate profile requires PSI metric to be enabled for the worker nodes (psi=1 kernel parameter). Directly check it for the node that is executing the operator pod, looking if /proc/pressure/ is available inside the pod, and report it with a condition.
Check it for all the worker nodes and report it also with an alert.

Fixes: https://issues.redhat.com/browse/WRKLDS-1673

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"context"
 	"fmt"
+	"os"
 	"slices"
 	"strconv"
 	"strings"
@@ -67,6 +68,8 @@ import (
 
 const DefaultImage = "quay.io/openshift/origin-descheduler:latest"
 const kubeVirtShedulableLabelSelector = "kubevirt.io/schedulable=true"
+const psiPath = "/proc/pressure/"
+const EXPERIMENTAL_DISABLE_PSI_CHECK = "EXPERIMENTAL_DISABLE_PSI_CHECK"
 
 // deschedulerCommand provides descheduler command with policyconfigfile mounted as volume and log-level for backwards
 // compatibility with 3.11
@@ -88,6 +91,7 @@ type TargetConfigReconciler struct {
 	namespaceLister          corev1listers.NamespaceLister
 	nodeLister               corev1listers.NodeLister
 	cache                    resourceapply.ResourceCache
+	psiPath                  string
 }
 
 func NewTargetConfigReconciler(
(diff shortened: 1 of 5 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -1594,6 +1630,18 @@ func (c *TargetConfigReconciler) isKubeVirtDeployed() (bool, error) {
 	return false, nil
 }
 
+func (c *TargetConfigReconciler) isPSIenabled() (bool, error) {
+	_, err := os.Stat(c.psiPath)
```
ingvagabund, line 1634: Even though the [operator deployment](https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/manifests/cluster-kube-descheduler-operator.clusterserviceversion.yaml#L347) does not currently tolerate control-plane/master taints it might be extended to do so in the future. When a PSI is enabled are non-worker nodes/container running on non-worker nodes expected to have `/proc/pressure` present too?

tiraboschi: PSI is supposed to be enabled via a {{machineConfig}} object. It only depends on the label selector set there by the cluster admin.

test/e2e/operator_test.go
```diff
@@ -886,3 +942,43 @@ func tryRemovingTaintWithExpectedSuccess(ctx context.Context, t *testing.T, clie
...
+		if v.Name != fakePSIVolume {
+			volumes = append(volumes, v)
+		}
+	}
+	for _, vm := range operatorDeployment.Spec.Template.Spec.Containers[0].VolumeMounts {
+		if vm.Name != fakePSIVolume {
+			volumeMounts = append(volumeMounts, vm)
+		}
+	}
+	return updateDeploymentAndRetryOnConflicts(ctx, kubeClient, operatorDeployment, metav1.UpdateOptions{})
```
ingvagabund, line 983: Neither of `volumes` and `volumeMounts` seems to be put back to `operatorDeployment`.

tiraboschi: Unfortunately we can't really mount anything under `/proc`.
Let's use an env variable to override it for testing purposes.

pkg/operator/target_config_reconciler.go
```diff
@@ -67,6 +68,8 @@ import (
 
 const DefaultImage = "quay.io/openshift/origin-descheduler:latest"
 const kubeVirtShedulableLabelSelector = "kubevirt.io/schedulable=true"
+const defaultPSIPath = "/proc/pressure/"
+const psiPathEnv = "PSI_PATH"
```
ingvagabund, line 72: New env means new API. If anything let's make it a boolean env. E.g. "DISABLE_PSI_CHECK" to limit the scope.

tiraboschi: Done

PR #510, opened 2025-04-22
[RelieveAndMigrate] provide defaults for DevDeviationThresholds and DevActualUtilizationProfile

tiraboschi:
Provide sane defaults for DevDeviationThresholds and
DevActualUtilizationProfile with the KubeVirtRelieveAndMigrate
profile.


pkg/operator/target_config_reconciler.go
```diff
@@ -1035,17 +1055,32 @@ func relieveAndMigrateProfile(profileCustomizations *deschedulerv1.ProfileCustom
 		setExcludedNamespacesForLowNodeUtilizationPlugin(profile.PluginConfigs[0].Args.Object.(*nodeutilization.LowNodeUtilizationArgs), includedNamespaces, excludedNamespaces, protectedNamespaces)
 	}
 
-	lowThreshold, highThreshold, err := getLowNodeUtilizationThresholds(profileCustomizations, false)
+	resourceNames := []v1.ResourceName{v1.ResourceCPU, v1.ResourceMemory, v1.ResourcePods}
```
ingvagabund, line 1058: This line can be dropped is it's always changed to `resourceNames = []v1.ResourceName{nodeutilization.MetricResource}`.

pkg/operator/target_config_reconciler.go
```diff
@@ -1035,17 +1055,32 @@ func relieveAndMigrateProfile(profileCustomizations *deschedulerv1.ProfileCustom
...
+	}
+	resourceNames = []v1.ResourceName{nodeutilization.MetricResource}
 
 	if profileCustomizations != nil {
 		// enable deviation
-		args.UseDeviationThresholds = profileCustomizations.DevDeviationThresholds != nil && *profileCustomizations.DevDeviationThresholds != ""
+		if profileCustomizations.DevDeviationThresholds != nil {
+			args.UseDeviationThresholds = *profileCustomizations.DevDeviationThresholds != ""
+		}
+		if profileCustomizations.DevLowNodeUtilizationThresholds != nil {
```
ingvagabund, line 1081: The code needs to error when both `DevDeviationThresholds` and `DevLowNodeUtilizationThresholds` are non-nil:
```go
if profileCustomizations.DevDeviationThresholds != nil && profileCustomizations.DevLowNodeUtilizationThresholds != nil {
	return nil, fmt.Errorf("only one of DevLowNodeUtilizationThresholds and DevDeviationThresholds customizations can be configured simultaneously")
}
args.UseDeviationThresholds = profileCustomizations.DevDeviationThresholds != nil && *profileCustomizations.DevDeviationThresholds != ""
```

pkg/operator/target_config_reconciler.go
```diff
@@ -1066,6 +1101,10 @@ func relieveAndMigrateProfile(profileCustomizations *deschedulerv1.ProfileCustom
 		}
 	}
 
+	lowThreshold, highThreshold, err := getRelieveAndMigrateThresholds(profileCustomizations, args.UseDeviationThresholds)
```
ingvagabund, line 1104: You could probably keep the original `getLowNodeUtilizationThresholds` around, set both `lowThreshold` and `highThreshold` to the default asymmetric `0` and `10` and invoke `getLowNodeUtilizationThresholds` only when `profileCustomizations != nil`.

pkg/operator/target_config_reconciler.go
tiraboschi: done

pkg/operator/target_config_reconciler.go
tiraboschi: `getRelieveAndMigrateThresholds` is internally doing that

PR #519, opened 2025-05-12
Add an alert and an condition for RelieveAndMigrate without PSI

ingvagabund:
RelieveAndMigrate profile requires PSI metric to be enabled for the worker nodes (psi=1 kernel parameter). Directly check it for the node that is executing the operator pod, looking if /proc/pressure/ is available inside the pod, and report it with a condition.
Check it for all the worker nodes and report it also with an alert.

Porting https://github.com/openshift/cluster-kube-descheduler-operator/pull/507 to 4.19.


ingvagabund:
Overriding as this was merge before the branching. Thus, the CI would pass anyway.

PR #563, opened 2025-05-27
OCPBUGS-57393: always configure metricsProviders with RelieveAndMigrate

tiraboschi:
RelieveAndMigrate is now (after #510) trying to consume by default a Prometheous query so metricsProviders should always appear in the descheduler configuration when RelieveAndMigrate profile is in use, even without any profileCustomizations.

Fixes: https://issues.redhat.com/browse/WRKLDS-1714

### DIFF
--- a/pkg/operator/target_config_reconciler_test.go
+++ b/pkg/operator/target_config_reconciler_test.go
@@ -286,6 +286,20 @@ func TestManageConfigMap(t *testing.T) {
 				TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: "ConfigMap"},
 				Data:     map[string]string{"policy.yaml": string(bindata.MustAsset("assets/relieveAndMigrateDefaults.yaml"))},
 			},
+			routes: []runtime.Object{
+				&routev1.Route{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "openshift-monitoring",
+						Name:      "prometheus-k8s",
+					},
+					Status: routev1.RouteStatus{Ingress: []routev1.RouteIngress{
+						{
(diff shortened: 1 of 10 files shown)


pkg/operator/target_config_reconciler_test.go
```diff
@@ -899,6 +1081,20 @@ func TestManageConfigMap(t *testing.T) {
...
+						Name:      "prometheus-k8s",
+					},
+					Status: routev1.RouteStatus{Ingress: []routev1.RouteIngress{
+						{
+							Host: "prometheus-k8s-openshift-monitoring.apps.example.com",
+						},
+					},
+					},
+				},
+			},
```
ricardomaraschini, line 1097: Shouldn't we add a case for "RelieveAndMigrate" without the prometheus route present ?

pkg/operator/target_config_reconciler_test.go
```diff
@@ -1530,9 +1741,15 @@ func TestSync(t *testing.T) {
 							},
 						},
 					},
+					&corev1.Namespace{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:   operatorclient.OperatorNamespace,
+							Labels: map[string]string{operatorclient.OpenshiftClusterMonitoringLabelKey: operatorclient.OpenshiftClusterMonitoringLabelValue},
```
ricardomaraschini, line 1747: Just for learning: why do we require the operator namespace to be labelled with "monitoring" ?

ricardomaraschini: There is a lovely typo in `checkNamepsaceMonitoringLabel()` :-)

tiraboschi: Done, thanks.

PR #614, opened 2025-06-26
[RelieveAndMigrate] Default opt-out for virt-launcher pods

tiraboschi:
The annotation `descheduler.alpha.kubernetes.io/evict` can be added to a Pod to explicitly mark it as eligible for eviction by the descheduler.
However, the descheduler only checks for the presence of the annotation key but its value is ignored. As a result, both `descheduler.alpha.kubernetes.io/evict=true` and `descheduler.alpha.kubernetes.io/evict=false` have the same effect: the pod becomes evictable.

This behavior prevents unprivileged users from opting their pods out of descheduler actions, which could otherwise defeat policy enforcement.
E.g: plugins that evict pods not complying with security or organizational rules.

At the same time, the DefaultEvictor plugin allows admins to scope eviction using label selectors.
To ensure that virtual machines are not evicted unintentionally when not expected, this change makes the descheduler exclude by default all pods labeled `kubevirt.io=virt-launcher` when the `RelieveAndMigrate` plugin is enabled.

VM owners can still opt their individual VMs in or out by adding or removing the `descheduler.alpha.kubernetes.io/evict` annotation but this control is limited to virt-launcher pods (that are off by default due to the label selector), avoiding unintended effects on unrelated workloads.

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -68,6 +68,8 @@ import (
 
 const DefaultImage = "quay.io/openshift/origin-descheduler:latest"
 const kubeVirtShedulableLabelSelector = "kubevirt.io/schedulable=true"
+const kubeVirtLabelSelectorKey = "kubevirt.io"
+const kubeVirtLabelVirtLauncherPod = "virt-launcher"
 const psiPath = "/proc/pressure/"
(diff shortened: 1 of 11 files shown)


ingvagabund:
Waiting for https://github.com/openshift/descheduler/pull/208

PR #642, opened 2025-08-04
[RelieveAndMigrate] configure Mandatory noEvictionPolicy

tiraboschi:
Confiogure the defaultEvictor plugin with
mandatory noEvictionPolicy so that
descheduler.alpha.kubernetes.io/prefer-no-eviction annotation will assume a mandatory value.
This can be used to opt-out selected KubeVirt VMs
without the need to reboot them.


ingvagabund:
@tiraboschi can you also mentions the use of the new no-policy/annotation in the readme at https://github.com/openshift/cluster-kube-descheduler-operator?tab=readme-ov-file#devkubevirtrelieveandmigrate?

PR #1744, opened 2026-04-17
KubeVirtRelieveAndMigrate: add stable scoring pipeline and eviction cooldown to reduce descheduling churn

tiraboschi:
This PR introduces two improvements to the Prometheus-based node scoring used by the descheduler, aimed at reducing instability and spurious eviction loops.

### Stable per-dimension noise filtering (p66/5m) 
Rather than computing the actuation priority from instantaneous 1-minute averages, each positive-deviation dimension (CPU utilization, CPU pressure, memory utilization, memory pressure) is now filtered independently through `quantile_over_time(0.66, ...[5m])` before being combined into the Euclidean distance. Applying the quantile per-dimension, rather than to the final score, is important because the Euclidean distance squares each input: filtering after squaring leaves transient spikes partially visible in the combined result. Per-dimension noise is also often asynchronous (a CPU spike in one interval, a memory spike in another); filtering independently prevents each spike from contributing to the distance at all. The existing :avg1m chain is preserved for comparison and debugging.                                                                                                                 
                  
### Eviction cooldown multiplier
A cooldown mechanism suppresses the actuation priority of nodes that have been recently evicted from, to prevent the descheduler from repeatedly targeting the same node before the effects of prior evictions have settled. The suppression is proportional to the number of successful `LowNodeUtilization` evictions in the past 15 minutes (linear decay, 25% per eviction), and decays naturally as old evictions leave the sliding window.
Known limitation: the cooldown is applied to the eviction source node. Ideally it would also dampen the score of the receiving node, but the descheduler is only involved in the eviction act itself while pod placement is entirely up to the scheduler, which the descheduler has no visibility into.
Requires: https://github.com/kubernetes-sigs/descheduler/pull/1856

### DIFF
--- a/bindata/assets/kube-descheduler/prometheusrule.yaml
+++ b/bindata/assets/kube-descheduler/prometheusrule.yaml
@@ -148,3 +148,29 @@ spec:
               3 * descheduler:node:ideal_point_positive_distance:p66_5m,
               1.0
             )
+
+        # Track successful eviction by LowNodeUtilization strategy count per node in the last 15 minutes
+        - record: descheduler:node:eviction_count:15m
+          expr: |-
+            label_replace(
+              sum by (node) (increase(descheduler_pods_evicted_total{strategy="LowNodeUtilization", result="success"}[15m])),
+              'instance', "$1", 'node', '(.+)'
+            ) or on (instance)
+            descheduler:nodeutilization:cpu:avg1m * 0
+
+        # Calculate the Dampening Factor (Multiplier)
+        # We use a linear decay: each eviction reduces the score by 25%.
+        # 4 evictions in 15m will effectively "mute" the node (multiplier close to 0).
+        - record: descheduler:node:cooldown_multiplier:15m
+          expr: |-
+            clamp_min(
+              1 - (descheduler:node:eviction_count:15m * 0.25),
+              0.01
+            )
+
+        # Actuation Priority: Stable Distance x Cooldown
+        # If the node was recently touched, the distance is suppressed.
+        - record: descheduler:node:actuation_priority:p66_5m
+          expr: |-
+            descheduler:node:linear_amplified_ideal_point_positive_distance:k3:p66_5m
+            * on(instance) descheduler:node:cooldown_multiplier:15m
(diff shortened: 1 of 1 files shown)


bindata/assets/kube-descheduler/prometheusrule.yaml
```diff
@@ -112,3 +112,65 @@ spec:
               3 * descheduler:node:ideal_point_positive_distance:avg1m,
               1.0
             )
+
+        # Stable per-dimension deviations: p66 over 5m
```
fabiand, line 116: how about p80? 80% seems to be a common threshold

bindata/assets/kube-descheduler/prometheusrule.yaml
```diff
@@ -112,3 +112,65 @@ spec:
...
+        - record: descheduler:node:eviction_count:15m
+          expr: |-
+            label_replace(
+              sum by (node) (increase(descheduler_pods_evicted_total{strategy="LowNodeUtilization", result="success"}[15m])),
+              'instance', "$1", 'node', '(.+)'
+            ) or on (instance)
+            descheduler:nodeutilization:cpu:avg1m * 0
+
+        # Calculate the Dampening Factor (Multiplier)
+        # We use a linear decay: each eviction reduces the score by 25%.
```
fabiand, line 162: So it is not reducing by 25%, because this owuld never converge, so we are saying it is a budget of 4 that we allow?

fabiand: Ah, I see the clamping is leading to the convergence.

bindata/assets/kube-descheduler/prometheusrule.yaml
tiraboschi: This is a quantile meaning the value that was >= then 66% of recorded data points over the last 5-minute window.
p50 is the median, p66 is the "worse-than-average but not outlier" value, p80 will capture shorter bursts of pressure or utilization that the p66 would ignore. p99 will capture all the short-lived spikes.

bindata/assets/kube-descheduler/prometheusrule.yaml
tiraboschi: It's an eviction penalty, not a simple boolean gate due to a budget of evictions over the time.
Even the first eviction will reduce the score of the node by 25%. So the second eviction will be triggered only if, even with the 25% penalization, the score of the node is still so high to classify it as overutilized and so on.

PR #2062, opened 2026-06-24
Fix protected namespaces not refreshing after operator start

tiraboschi:
The list of protected namespaces (`openshift-*`) was built once in `NewTargetConfigReconciler` via a direct API call and cached in the struct field for the lifetime of the process. Namespaces created after the operator started (e.g. `openshift-cnv` from a late CNV install) were never added to the exclusion list, leaving them unprotected from descheduling until the operator pod restarted.

Replace the cached `protectedNamespaces` field with a `getProtectedNamespaces()` method that queries the namespace informer cache on each reconciliation. The namespace informer was already present and synced, so the performance impact is negligible: only the derived list of protected namespaces was unnecessarily cached. The list is now sorted for deterministic output.

Add `TestGetProtectedNamespacesRefresh` to explicitly verify that namespaces created after initialization are picked up.

Fixes: [CNV-89284](https://redhat.atlassian.net/browse/CNV-89284)

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -93,7 +93,6 @@ type TargetConfigReconciler struct {
 	dynamicClient            dynamic.Interface
 	eventRecorder            events.Recorder
 	queue                    workqueue.RateLimitingInterface
-	protectedNamespaces      []string
 	configSchedulerLister    configlistersv1.SchedulerLister
 	routeRouteLister         routelistersv1.RouteLister
 	namespaceLister          corev1listers.NamespaceLister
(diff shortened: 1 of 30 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -156,6 +141,21 @@ func NewTargetConfigReconciler(
 	return c
 }
 
+func (c *TargetConfigReconciler) getProtectedNamespaces() ([]string, error) {
+	allNamespaces, err := c.namespaceLister.List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("failed to list namespaces: %w", err)
+	}
+	protectedNamespaces := []string{"kube-system", "hypershift", "openshift"}
```
ingvagabund, line 149: Alternatively, we can patch o/descheduler and always exclude protected openshift namespaces. In `DefaultEvictor.Filter`. We already know we don't ever want to evict from the protected namespaces. Also, this is openshift specific. We can carry the patch indefinitely.

PR #2064, opened 2026-06-25
no-jira: reduce rbac permissions

ingvagabund:
Fill in CLAUDE.md with intel about the RBAC for the operator and the operands. Also, while at it reducing the permissions granted.


ingvagabund:
/retest-required

`make test-e2e` completes succesfuly locally.

